PR #7, opened 2019-08-12
Yogee Token Image added to Trust wallet

cyberience:
This is a Smart contract token for Yogee Coin. Correct previous mistake of pasting the wrong contract address. now fixed.


cyberience:
Whats the best solution to this problem? should I just delete the folder, or do a fresh fork?

PR #20, opened 2019-08-13
Image for Smart Pharma Token and Ethereum address.

cyberience:
SMTP Token asset.


cyberience:
I seem to be having a problem placing this asset. the token address is this: https://etherscan.io/address/0x02f28544c89b2d704eb0d1f9f4fc60d688ad8959
So it is legitimate, also lower case, and a valid PNG file. Any suggestions?

kolya182:
@cyberience You everything right, thanks for PR.

PR #86, opened 2019-08-26
AEROTOKEN (AET)- first self inflationary token

aerotoken:
aerotoken
AEROTOKEN (AET) is an inflationary crypto-token, operating on the Ethereum blockchain (ERC-20 standard token) which can be used as a currency, a store of value or an investment.
Each time an AET is transferred, 1% of the transaction is increased.


aerotoken:
aerotoken
AEROTOKEN (AET) is an inflationary crypto-token, operating on the Ethereum blockchain (ERC-20 standard token) which can be used as a currency, a store of value or an investment.
Each time an AET is transferred, 1% of the transaction is increased.

PR #88, opened 2019-08-27
Add OROM

OromEx:
Thank you for adding Orom Token
- [x] Filename in the lowercase formar
- [x] Image format: png
- [x] Image size: 256 by 256 px


hewigovens:
tracked here https://github.com/trustwallet/assets/pull/93

PR #119, opened 2019-09-01
Update list.json to include POS Bakerz

EdouardLvdl:


vikmeup:
@EdouardLvdl can you also add an image too, right here: https://github.com/trustwallet/assets/tree/master/blockchains/cosmos/validators/assets

PR #162, opened 2019-09-10
Add SparkPool to validator lists of Cosmos

yulidai:


yulidai:
It doesn't seem to be my problem, but there is a problem with the existing folder name.

PR #230, opened 2019-09-17
Add Precatorio RMDSP01

saintclair:
Add Precatorio RMDSP01


kolya182:
Why would you want to replace ethereum logo ?

saintclair:
Fixed

kolya182:
@saintclair Thanks, the reason i asked, i noticed a few other contributors also did the same, is anything i can change in guidelines to make process more clear ?

saintclair:
I understand, in the folder structure (tree) the logo.png could have another name in the info because the user thinks it has to put in both places.

kolya182:
> I understand, in the folder structure (tree) the logo.png could have another name in the info because the user thinks it has to put in both places.

Hope this help https://github.com/trustwallet/assets/commit/826862f0d90989c1f9c092e176ad4c64ce143443

saintclair:
@kolya182 Now I guess so, it's clear, taking advantage of when you think it's available in my app?

PR #251, opened 2019-09-20
Update list.json for TRON, added InfStones

sili-infstones:
Update list.json for TRON, added InfStones


kolya182:
And validator image too in lowercase register, thank you

PR #267, opened 2019-09-23
Add Hatch (HATCH)

hatchpay:


hatchpay:
Please write what is the reason?

kolya182:
> Please write what is the reason?

We don't support it.

PR #268, opened 2019-09-23
Update list.json

tezosbc:
Entry for TezosBC added.


kolya182:
Please add validator image (lowercase register)

PR #269, opened 2019-09-24
Update list.json

unifi-protocol:


blockchains/tron/validators/list.json
```diff
@@ -2,7 +2,7 @@
     {
         "id": "TGzz8gjYiYRqpfmDwnLxfgPuLVNmpCswVp",
         "name": "Sesameseed",
-        "description": "Sesameseed is a blockchain community providing fair and transparent representation in delegated governance.",
+        "description": "Sesameseed is a blockchain community providing fair and transparent representation in delegated governance by rewarding Voters for their participation on Tron and Ontology.",
```
kolya182:
```suggestion
        "description": "Sesameseed is a blockchain community providing fair and transparent representation in delegated governance by rewarding Voters for their participation.",
```

kolya182: Sounds like it rewords on both chains by participating on Tron only

sesameseed-org: Ahh...

I understand. Technically it does reward on both chains by participating on Tron Only. The token "SEED" we reward the user in can be redeemed for the representative node rewards on both chains. (e.g. 1 seed = X TRX and Y ONG)

But your change is great, we can use that until other nodes become active

kolya182: Once we support ONG staking ,we can came back and change it

PR #270, opened 2019-09-24
Add BNB folder as replacement for Binance

kolya182:


hewigovens:
I suggest to not change this, old clients would break, only change name in info should be enough

kolya182:
> I suggest to not change this, old clients would break, only change name in info should be enough

I will keep old folder `binance` for a while, and new `bnb` will be adopted i next releases

PR #359, opened 2019-10-08
added waves validator wavescap

WavescapTorben:


vikmeup:
can you lowercase name of the image please?

PR #400, opened 2019-10-15
TBT Token added

TBT-VC:
TBT Contract Address:0x62a7dA217d6a55A983455e5fd6C4a025aC786879
TBT Supply: 100,000,000,000
TBT Decimal: 6
TBT symbol: TBT


TBT-VC:
folder name is a contract address and it can be in lower case

PR #417, opened 2019-10-16
Added token logo

lcraciun25:
Added TSB logo image


kolya182:
How about contract address ?

PR #424, opened 2019-10-17
Everstake add information in cosmos and tezos

everstake:


everstake:
Pull request with logos and information about Everstake in cosmos and tezos failed. How to resolv this problem?

PR #426, opened 2019-10-17
Create  ANEX

Vahe327:


kolya182:
We don;t support token info at the moment, will add this later.

PR #437, opened 2019-10-18
Lab

ghost:


kolya182:
@mg-8 As i mentioned to you many times before, when you make new PR, you need update your fork, other-ways to upload same images over again and that what it cause merging conflict as a result delayed PR merge. Fixing for you last time, after i will just close. Lets' respect each other time.

PR #548, opened 2019-10-28
please update dark matter token 1002575 to tron trc-10 i already set it up

dorobiala:
I hope this is what I am supposed to do


kolya182:
Added here https://github.com/trustwallet/assets/commit/7cf171c54ad4be59f88edf588c17e35885d00009

PR #591, opened 2019-11-09
almg token added

dexdaq:
almg token added


dexdaq:
Hi again

Still Almeegram doesn't have any logos at trust wallet.Can you please
inform us .Beaceuse my other request for dexdaq(Ddq) token has been
completed.But for Almeegram still the same

PR #600, opened 2019-11-11
VTHO is a smart contract on the blockchain

ArtemGoryunov:


vikmeup:
@kolya182 we need an image for both, checksum and nonchecksum? can you make that happen to support old versions and current one?

PR #617, opened 2019-11-13
Add Paz files via upload

pazcoinfoundation:
[x] image format ‘png’
[x] image size : ‘256x256 px’


kolya182:
Why two images ?. need only one in checksum

PR #624, opened 2019-11-15
Added a new TTK Token

klienlinll:
Adding a new TTK Token on ThunderCore


klienlinll:
Fail to pass address checksum

PR #680, opened 2019-11-21
Adding Wanchain validator folder and adding Anonstake validator info

Anonstake:


kolya182:
FYI: We don't support staking on Wanchain., but you will be the first in the list once we do.

PR #699, opened 2019-11-25
Our very first experience

Atlantis-eFork:
Hey guys, the is our very first time.

we have a large community that uses your wallet daily, so we would be very thankful if you guys would add us. 

Thanks in Advance

Atlantis Team


kolya182:
This week we enforcing a new contribution rules, that will require you posting backlink to Trust Wallet website on your website. more details will be provided in couple of days.

kolya182:
You adding image to address, but we support only images for contract.

PR #700, opened 2019-11-25
Removal of UNV token

Dan-UVC:
No longer supported - replaces by UVC


kolya182:
I see this is not even contract address, removed. Thanks

PR #755, opened 2019-12-07
Add CR Coin (CRC)

Gabo1122:
CR Coin is Costa Rica's first utility token, powered by Blockchain Costa Rica


Gabo1122:
CR Coin is Costa Rica's utility token

PR #791, opened 2019-12-11
Add OXT Token

nhandler:
This change adds the Orchid OXT ERC-20 token.


nhandler:
The tests were failing due to incorrectly named duplicate directories that existed in the repository. I've gone ahead and cleaned these up so that the tests are able to properly run. Please let me know if you prefer that these extra commits be moved to a separate Pull Request and/or squashed (GitHub also allows them to be squashed at merge time through their UI ).

PR #801, opened 2019-12-12
remove wrong coin

Pantani:
The first one is right for BTC
https://github.com/trustwallet/assets/blob/master/pricing/coinmarketcap/mapping.json/?#L3
the second one is wrong, this is for USDT (id 825)
https://github.com/trustwallet/assets/blob/master/pricing/coinmarketcap/mapping.json/?#L396


vikmeup:
I think we need to fix code that cause this issue: https://github.com/trustwallet/assets/pull/740/files#diff-2236ca83d9d252261bded20ec411952aR202, otherwise it will happen again

kolya182:
Let me modify script and add more tests

Pantani:
yes, this looks like a merge conflict problem

PR #825, opened 2019-12-14
Add Tezos validator - HappyTezos

andrewpaulicek:


kolya182:
Hi @andrewpaulicek ca, you find me in telegram t.me/mykola

PR #826, opened 2019-12-14
Uploading Onyx logos

ghost:


kolya182:
We will be adding prices from Coingecko in couple of week for your tokens

PR #831, opened 2019-12-15
Update CMME logo

cmmeio:


kolya182:
@cmmeio You completely changing icon image, do you have link to where i can check if this is correct logo for this token ?

PR #853, opened 2019-12-17
Add WavesAssist to Trust Wallet

sanderwaves:


kolya182:
@sanderwaves Are you adding these two on behave of validators ?. Please find me in telegram t.me/mykola

PR #855, opened 2019-12-17
Added P2P Validator to Tezos

vadimicus:
Happy to be part of the Trust platform


kolya182:
@vadimicus Website seems to be not working https://www.host-tracker.com/InstantCheck/2/cdf8d176-0821-ea11-a5ff-0003ff7328cc/ please find me in telegram t.me/mykola

vadimicus:
We where under the maintenance, and it was just front. All Validators are alive with 100% uptime.

PR #860, opened 2019-12-17
Update cmc map

kolya182:
What changed
- Added new ERC20 contracts and removed no longer available 
- Updated some contract
- Preserve if coin id order for tickers with same id


Pantani:
Do we already have a test for duplicate records?

kolya182:
> Do we already have a test for duplicate records?
Not exactly to check if objects deeply equal (tested this case and jest hang when looping overs elemetns) but we have other test to check  duplicates. But in general BA should ignore duplicates

PR #871, opened 2019-12-19
Add logo via upload

ceocharles:


leecharles:
The logo does not display on TrustWallet.

ceocharles:
The logo is visible on the iPhone, but not on the Android phone. And on the iPhone, even though it is registered as GoChain, it is displayed as ERC20, not G020.

PR #878, opened 2019-12-20
Upload NRVCOIN logo

AnuningMDT01:


kolya182:
You don't need make changes to multiple images, here is how it should be donehttps://github.com/trustwallet/assets/commit/88732a533f7a327ed5bacb9a8f612747687796f8

PR #889, opened 2019-12-21
Validity (VLDY) image integration

deoidh:
Address: 0x904da022AbCF44EbA68d4255914141298a7F7307
Webiste: vldy.org


samgos:
Seems that IND token didn't checksum their address hence the build fail, 
will leave it to the repo operator to solve.

PR #910, opened 2019-12-25
Add payout percents for Tron and Tezos validators

dpereskokov:
Source info for Tron validators payout percents:
```
- https://www.tron2moon.com/calculator.php
- https://tron.live/tronsr/
```

Source info for Tezos validators payout percents:
```
- https://tezos.fish/ru/leaderboard/all
```

Several validators couldn't be found in the sources and kept payout as 0%

Fixes: #909


vikmeup:
Can we not use 0%? There is some validators that pay out and it set to 0%

PR #930, opened 2019-12-31
Add Baking Tacos tezos validator

utdrmac:
This PR adds Baking Tacos as a validator to Tezos. Baking Tacos is listed on Ledger Live, and Atomic Wallet. They should be listed here too.


kolya182:
Hi @utdrmac Are you adding on behalf of validator ?.  If yes, please contact me in telegram (see bio).

PR #947, opened 2020-01-03
Request to Remove Old Duplicate Token Name

cbucks:
We would like to completely remove the listing of our previous old duplicate token so that it will not appear on the search results.

We need to remove this because our token holders are confused.

I removed the address from the whitelist.json and added it into the blacklist.json file.

Please advise if this is the correct way of removing a token from the search results.

Thank you and have a nice day.


kolya182:
I will fix this today

PR #949, opened 2020-01-03
upload token logo

aburaza:


aburaza:
Upload logo for payusd token new .
Please help me update it. Thanks so much

PR #954, opened 2020-01-04
add "melea ◮👁◭" Cosmos Validator 

iammelea:


kolya182:
@iammelea  Can you find me in telegram, see bio

PR #958, opened 2020-01-04
Please add Blockscape.validator

BlockscapeLab:


kolya182:
@BlockscapeLab Can you find me in telegram, see bio

PR #960, opened 2020-01-05
add Staking Fund

1xCL:


kolya182:
@winslyn can you find my in telegram, see bio

PR #963, opened 2020-01-05
Add Cosmos validator in3s.com

davepuchyr:


kolya182:
@davepuchyr Please move to the bottom of the list, and find me in telegram, see bio.

PR #977, opened 2020-01-08
Add files via upload

pyrrhos-gold:
Image size is 256px x 256px
image is in .png format
file name is lowercase


vikmeup:
Image name should be checksum, please re-upload the file when ready

PR #979, opened 2020-01-08
ADD GNTOKEN to TRUST WALLET

tokyo786:


vikmeup:
What is the purpose of the token?

lavimalik:
To be used for transactions in GN holdings.

PR #989, opened 2020-01-10
Added LootUp Token

autotheos:
Added the LootUp Token with the proper size png file


kolya182:
Just add logo, no need to run list genration

autotheos:
Sorry what do you mean by list generation?

Issue #1017, opened 2020-01-14
Script does not check if address is checksum


vikmeup:
https://github.com/trustwallet/assets/pull/1016

sidko:
I see 4 of these addresses seem to have failed validation: 
```
0xa4f9cec920ca520a7feb2c3a63050e08967bc111
0xb19977b23b40d59254b8f5a0bdb67076b4fa0bc6
0xcf0425a35f1fa900bb347cb996d3094d1df3cd48
0xe6710e0cda178f3d921f456902707b0d4c4a332b
```
Should I re-submit them with the right checksum format and the same image?

vikmeup:
@sidko yes please, use `git mv` to checksum those.

we will add script to CI those values

sidko:
@vikmeup just a heads up, the diff in git may not show this correctly because the change is only for the folder case. Also, I used `git mv` twice because git doesn't like me changing only the case of the folder (see https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv).

kolya182:
> @vikmeup just a heads up, the diff in git may not show this correctly because the change is only for the folder case. Also, I used `git mv` twice because git doesn't like me changing only the case of the folder (see https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv).

`npm run checksum-erc20` script already there

vikmeup:
you need to run this automatically and commit changes on PR

PR #1025, opened 2020-01-14
Change folder names to checksum

sidko:
See #1017


vikmeup:
Here is an example: https://github.com/trustwallet/assets/commit/c3333d30230144e6a2a34869f495abebc639ad9d, it show my changes correctly for folder rename

sidko:
The commit above shows the rename of these exact 4 folders. Does it mean the rename has already taken place?

vikmeup:
Yes, I did rename it yesterday. I just created an issue that CI would verify addresses on each PR.

sidko:
Ah amazing, thanks for being so on top of this! I think this PR can be closed now - the purpose was just to make sure the rename went through. I'll add more logos for social money in another PR for your review.

PR #1035, opened 2020-01-16
Made logo.png transparent

mccallj57:


kolya182:
Made logo up to size standard

kolya182:
Find better logo quality,

PR #1037, opened 2020-01-16
New logo.png

mccallj57:
The gears are to represent TIME in a watch


mccallj57:
Sorry, I am new to the process, but was wondering if you could provide some guidance as to why the commit was closed without merging?

PR #1040, opened 2020-01-17
Add SG10006

larryfranky:
File is in lowecar and has a .png file, image size is 256px by 256px


kolya182:
You must be reading outdated requirements, added your logo.

PR #1068, opened 2020-01-22
Update kava logo

rhuairahrighairidh:
Looking to update the kava token logo to be a bit cleaner.
Image size is 256x256, transparent background, run through tinypng.


Sashane969 (commented):
blockchains/kava/info/logo.png

PR #1109, opened 2020-01-29
RealT Tokens

calebsheridan:


kolya182:
Cool stuff, what do you do with these tokens, curious to learn ?

calebsheridan:
This set of tokens each represent ownership in specific real estate properties which generate rental income. See https://realt.co/

PR #1153, opened 2020-02-04
Whitelist ECYT token

vcoolish:


kolya182:
Whitest is auto generated, i will take a look what is about this address

vcoolish:
Didn't know it's automatically

kolya182:
Token has not image as minimum requirement to get on whitelist

PR #1258, opened 2020-02-18
Add token info.json

zachzwei:
Adding token info and provided basic information about  top 35 token assets on CMC.


ismaelapariciojr (commented):
**@cryptomanz1 @ **

PR #1438, opened 2020-03-12
Ztar Icon Update

marco2142:
Ztar Icon Update 256x256 png format


marco2142:
i need update the icon

PR #1466, opened 2020-03-16
Solana: Updated logo and init validators

CriesofCarrots:
Update solana logo, and add empty validators file.

We will be fleshing out this file as we onboard validators over the next few days, but in the meantime, I am hoping the existence of the file will unblock our blockatlas integration.


CriesofCarrots:
The CI failures don't seem related to solana; I rebased on master in case I was missing a fix. Please let me know if I am missing something.

kolya182:
I will take a look

PR #1474, opened 2020-03-17
0x61D0ADfCF76225F3060eD9F72a5d74523c8eCabE.json

ghost:
obidua Provides Customers for retailers from his nearby location around 0-2 km that’s why retailer can easily deliver the item to customers doorstep within 10-30 minutes of rapid fast delivery. If retailers got an average of 5 to 10 orders per day than we provide him facility of a delivery Counter/Boy. We also catching up rural areas users through pickup and drop point facility to serve batter way.

logo


MznRahi:
logo

MznRahi:
logo

PR #1540, opened 2020-03-24
Update info

StakingShop:
Atm is full capacity


TezoShop:
hello guys, my baking capacity for new delegations at the moment is full. I ask you to suspend my service until I increase the capacity. thanks

kolya182:
Thanks for an update, we will need to implement automatic capacity check on https://github.com/trustwallet/blockatlas, any ideas how we can do that ?

TezoShop:
Hi, automatic capacity check in tezos does not make sense, as delegates can come and go at any time. The best option remains the opinion of the validator on the need to close the reception of delegations. This happens quite rarely with large validators. With small bakers like me, this can happen more often.

PR #1573, opened 2020-03-26
Add RIZIN Token Logo

rizintechnology:
Token Name: RIZIN
Symbol: RZN
Contract Address: 0xA13d570cea90D9Ae0AAe2C2BbA975645BC82ae85


zachzwei:
@kolya182 
This token does not appear when search via token name, ticker and even contract.
Can you check please?
Thanks.

PR #1629, opened 2020-03-30
Please Add this Token

ghost:
Tether DeFi (USDD) is called a stablecoin because it was originally designed to always be worth $1.00, maintaining $1.00 in reserves for each tether issued. Nevertheless, Tether DeFi (USDD) Limited states that owners of tethers have no contractual right, other legal claims, or guarantee that tethers will be redeemed or exchanged for dollars.


kolya182:
Stop spamming with this shitcoins

PR #1638, opened 2020-03-31
ETLX not appearing on search bar *BUG*

electrolitetoken:
ETLX not appearing on search bar *BUG*


kolya182:
I'm able to search by `ETXL` and `ETL`, what issue do you experience ?

PR #1669, opened 2020-04-02
Update ANT icon

sohkai:
See https://blog.aragon.org/aragon-brand-refresh/


sohkai:
@kolya182 Hey! Just wondering—it looks like the resize made the image _quite_ a lot larger! Should we run this through the optimizer again?

kolya182:
> @kolya182 Hey 👋! Just wondering—it looks like the resize made the image _quite_ a lot larger! Should we run this through the optimizer again?

Try it, but it did optimized proportionally to original image

sohkai:
Ran it through ImageOptim on 80% and got it to 70kb :). We might be able to do even a bit more—it was really nice that the old icon was only 17kb!

-------

Optimized version:

logo

kolya182:
Thanks, updated

PR #1721, opened 2020-04-06
Add ERC-20 XSGD Token Asset

liewjiajun:
Adding token logo for the ERC-20 XSGD token


liewjiajun:
Not sure why did test fail, please let me know what to do and I will make it right.

PR #1744, opened 2020-04-07
Lossless image Optimization

FlipFloop:


kolya182:
@FlipFloop Thank you for optimization,  could you make PR with a code that optimize images ?. We currently have Github image optimization in place, but as i see your is better. Closing for now.

FlipFloop:
with a code?

PR #1793, opened 2020-04-10
Add HETM

salehbancin:
Add HETM and Logo


kolya182:
@salehbancin Please pay the fee mentioned above, and we will merge PR.

PR #1840, opened 2020-04-12
InterGalactic Credits CREE

geopayme:
InterGalactic Credits (CREE) are used as a multi-purpose mediation instrument to interact with other decentralized interPlanetary Nodes.


kolya182:
> Is the error a system failure? is seems that this error happens 48% of the time? what is the resolution to this issue?

Solution is always simple as it can be, simply follow guidelines and do not upload lowercase and chacksumed version of contract. You need to delete lowercased version and paid the fee.

kolya182:
Added here https://github.com/trustwallet/assets/commit/8fa2a4fde1b58f3587347928dc089db3720d42af

Issue #1851, opened 2020-04-13
Setup Danger


vikmeup:
https://github.com/danger/danger

PR #1857, opened 2020-04-13
Add Danger

vikmeup:
Fixes: https://github.com/trustwallet/assets/issues/1851


.github/workflows/danger.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  danger:
+    runs-on: macos-latest
```
hewigovens, line 11: why macos? it's expensive

vikmeup: It’s free in public repos. But I can switch to Ubuntu

PR #1887, opened 2020-04-16
Update labels used by merge-fee-bot.

optout21:


catenocrypt:
Updated labels used by merge-fee-bot:
- when payment request message is placed, 'Payment Status: Requested' label created (will stay there)
- when payment has been executed, 'Payment Status: Paid' label is places.
I want to avoid removing labels, that's not transaction safe.

PR #1890, opened 2020-04-17
Updates pTokens pBTC

allemanfredi:
This PR contains:
* adding of __info.json__
* changing of __logo.png__


kolya182:
@allemanfredi You still need to pay the fee, and PR will be merged right away.

PR #1918, opened 2020-04-19
change token logo

bitcollarmax:
git add -A
git commit -m “Add <Bitcollar Max>”
git push origin <branch_name>


kolya182:
Updated

PR #1930, opened 2020-04-20
update

StakingShop:
Hello! i increased capacity of my baking address. please include my in the list of bakers. thanks

### DIFF
--- a/blockchains/tezos/validators/list.json
+++ b/blockchains/tezos/validators/list.json
@@ -282,7 +282,7 @@
             "commission": 3
         },
         "status": {
-            "disabled": true,
+            "disabled": false,
             "note:": "No more capacity"
         }
     },


TezoShop:
I received some letters in the mail about the need to pay 1000 TWT tokens to your team. Is this really so?

kolya182:
@TezoShop I see your max capacity almost reached, and we need improve baker experience by checking capacity dynamically rater hardcoding it here. Issue to track https://github.com/trustwallet/blockatlas/issues/1040

TezoShop:
@kolya182 I have increased capacity today and will be able to receive more than 50k xtz new delegations. I see your invention and I like it if the option will automatically remove the baker from the list when it is full of

kolya182:
@TezoShop Can you confirm if this address did not receive pay out due to maxed out capacity https://tezblock.io/account/tz1e6ftZCjNF7F5RBVcALbRw6E23txiLQiEd?tab=Delegations ?

TezoShop:
no it's not because of that. we set up our payment instrument in such a way that it would not make a payment if the balance at the address from which delegated 0 XTZ. if we send rewards to the address with 0 XTZ, we are forced to pay burn fee 0.257 XTZ for re-activation of this address. there are so many delegates who delegate with a small balance (0,0001 XTZ) and who transfer balances after few days to the exchange whose rewards are less than the amount of combustion. it is not profitable for a baker to sent awards at such addresses.

zachzwei:
The user was delegating for a month and has just moved his funds 2 days ago.
The 0 XTZ balance is just recent.
Never got any payouts.

TezoShop:
@zachzwei delegators in Tezos don’t receive any rewards for the first 7 cycles. That’s due to the protocol implementation. And then it takes 5 cycles for your rewards to become liquid before they can be distributed. To receive the first reward, the delegate needs to wait 34 days. he emptied his address at the wrong time.

TezoShop:
Let's say you delegate 5 XTZ, so your reward for the cycle will be approximately 0.00001 XTZ. if you empty your address every time before receiving a reward, then for 5 cycles the commission for burning will be more than 1 XTZ (0,257*5) at a time when only 0,00005 XTZ will be sent to your address. it is not profitable for bakers

zachzwei:
Thanks for the information. User advised me that he was waiting for a response from your guys to know why he did not get any payouts. Will relay this info to the user then.

TezoShop:
all right, I'll put it on the list for an exception

kolya182:
@TezoShop Thanks, lets keep in touch

TezoShop:
@kolya182 @zachzwei For example, I used the official faucet that sends a real drop of 0.01 XTZ to anyone. I requested for an address at which the balance is 0 xtz. if you look at the value "burned" in hash, it is 0.257. So the faucet paid for the burning and sent a drop 0,01 XTZ for this address 

https://tzstats.com/op7eK1nPj74Jo4Bw5XW6N3K5AuNsk73P23idDWVeGo3V9gJe7vk

TezoShop:
I paid him a reward, but I can’t send each such address a drop for burned fee first and then a reward - this is complicated and takes a lot of time and manual work.

PR #1942, opened 2020-04-21
add logo

whynoteasy:


kolya182:
@whynoteasy You don't need to create multiple pull request with the same content, you just need to pay for this one.

PR #2008, opened 2020-04-28
Add PALA INT COMM

zyjblockchain:
add pala token


zyjblockchain:
please merge my pr, thanks

PR #2025, opened 2020-05-04
Update CACHE icon

conscott:
Follow up to #2023 

Increased icon border so it looks better in the app


vikmeup:
@conscott just FYI: image might be cached on the iOS/Android up to 7 days, in the case that image is not changed, try a new device to see updated image.

PR #2032, opened 2020-05-05
Darphane $Chat Token Integration

ersanyakit:
Maximum PR age reached, auto-closing PR  (60 vs. 60 hrs)


ersanyakit:
Omg, its very interesting idea. I never seen before... :)

kolya182:
@ersanyakit PR ready to get merged, you just need to pay the fee.

PendelinP:
@kolya182 Why don't you simply support ETH payments? Your token is only listen on suspect/minor exchanges and BNB is a bit weird as well. I really don't get it why you make things so complicated?

vikmeup:
@PendelinP ETH is little trickier to implement, as it will require to create separate addresses for each pull request, much harder to manage and sweep funds into one wallet after.

PendelinP:
@vikmeup Thanks for the clarification. Would be great to have some more coins/some easier way to get this done :)

ersanyakit:
@kolya182 is there anyway to send ETH payments for this PR?

If its possible can u share with me your ETH address

Best Regards.

PR #2259, opened 2020-06-03
Update logo

frpalmer:
Update logo PLM


frpalmes:
The logo needs to be updated as well.

kolya182:
> The logo needs to be updated as well.

Just updated

PR #2308, opened 2020-06-09
Added Flama Token

ScorpioTM:
Flama is one of the first deflationary tokens in the market with enhanced futures like staking


ScorpioTM:
Thanks! I will be waiting.

PR #2371, opened 2020-06-14
add icon to monnos token

Abdoubody2000:


Abdoubody2000:
Hello
I reinstalled app and icon still not shown, Why ?

PR #2427, opened 2020-06-21
GetBlockCard.com TERN Crypto credit card.

kingassune:


kingassune:
Thank you

kaycoinsUSA:
Where is the credit card info? My Trust wallet is not updating. I have yet to get any wallets to work. I haven't paid myself. Bot you can pay me some BTC so I can pay my sponsors. We can just use this wallet if you can help me get a transaction that would be great. The largest btc you can do, then send me info on debit card.

PR #2431, opened 2020-06-21
EMIT is a Token Liquidity layer for a liquidity pool.

DrZedd42:


zachzwei:
The logo submission has already been verified and was just waiting on payment.
Noticed that you put in new logos and still used the wrong method.
Reverting the commits.

PR #2438, opened 2020-06-21
Different address 

kingassune:
Made a mistake


zachzwei:
Why is it the same token logo for 2 different contract address?

kingassune:
Made a mistake with the contract. We are only using the new one. 0x24283732C3DF91EE1e4354a3934C14B88B2d9a51

zachzwei:
Removed 0x9687f9dbebb76a2a469925cce797f3015d2c68fd.
Will merge once checks are done.

PR #2508, opened 2020-06-27
pyrabank private

pyrabank:


catenocrypt:
Please consult the contribution guidelines:
https://github.com/trustwallet/assets#contribution-steps

Log should be called logo.png, and I think the location is also not right, should be in the form `blockchains/ethereum/assets/0x19fFfd124CD9089E21026d10dA97f8cD6B442Bff`.

PR #2525, opened 2020-06-29
Add Cybercoin token

HelloMyDevWorld:
- [x] File name in the lowercase format
- [x] Image format: "png"
- [x] Image size: "256 by 256 px"

### DIFF
--- a/blockchains/ethereum/assets/0x4Ba012f6e411a1bE55b98E9E62C3A4ceb16eC88B/info.json
+++ b/blockchains/ethereum/assets/0x4Ba012f6e411a1bE55b98E9E62C3A4ceb16eC88B/info.json
@@ -1,0 +1,20 @@
+{
+    "name": "Cybercoin",
+    "website": "https://www.cybercoin.site",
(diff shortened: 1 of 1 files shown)


patsadow2:
Done. https://explorer.binance.org/tx/08E2A0DAEB215AAA5DF30D5757032855E5D4CA347DAFD906A0DB8E1BDB3D918C

I have additional question when i want to change something in the future i have to pay again ?

patsadow2:
I've got 2 questions:

I have additional question when i want to change something in the future i have to pay again ?

Does it make me also visible on uniswap ? Or do i have to add it itself somewhere ? But where ? :D

catenocrypt:
> When i want to change something in the future i have to pay again ?
> Does it make me also visible on uniswap ? Or do i have to add it itself somewhere ? But where ? :D

Our policy is that each PR requires new fee, even if it is change to an existing asset.  Thanks for your understanding.
Uniswap logos are independent of Trustwallet, it will not appear there automatically, I think.

PR #2532, opened 2020-06-30
Please add MARE on TrustWallet

soliditycdr:
Token Name	: Market Return
Symbol		: MARE
Decimals	: 8
Total Supply	: 1,000,000,000,000,000 MARE
Contract Address: 0x98eb850a587D908980328ED3Fde3e335C0cc4426
Original Logo: 
logo


soliditycdr:
Payment Done. Please check Memo.

catenocrypt:
@soliditycdr , are you sure this image will look OK as a relatively small logo?  Maybe unused border could be reduced.
Please confirm once more before us merging.

soliditycdr:
Hi @catenocrypt, sorry for late reply, was away due to family task. I just received this from my client and seems this is ok, I put main frame with 256x256, just make sure MARE appears on symbol. Thanks.

PR #2533, opened 2020-06-30
Add XAVIA token logo

xavia-token:
Add XAVIA token logo


xavia-token:
I have made the payment

PR #2540, opened 2020-07-01
Add KASH

oxmah:
Add KASH logo.


ogreports:
The payment has been sent.

PR #2741, opened 2020-07-14
bzrx and vbzrx tokens added

sneg55:


tomcbean:
Payment has been made

PR #2751, opened 2020-07-15
Add Havens Nook (HXN)

Havensnook:
Add Havens Nook (HXN) logo


Havensnook:
payment done 
Add logo HXN payment proof

PR #2768, opened 2020-07-17
Added Flapp Token

ScorpioTM:
Flapp is the token of both the present and the future: an ecosystem for developers and users that facilitates the use and creation of blockchain based dapps.
https://flapptoken.io/
https://etherscan.io/token/0xcfb72ed3647cc8e7fa52e4f121ecdabefc305e7f


ScorpioTM:
Payment done!

PR #2782, opened 2020-07-17
Add new coin pair to pricing

EnoRage:


acashcoin:
@kolya182  we paid , please we need to update the icon 
Thanks

Issue #2694, opened 2020-07-11
[WIP] Improve Validation


vikmeup:
Mission: Assets repo to be the most comprehensive and up-to-date coin (sh@tcoin) repository available, used by several projects.

Identified Problems:
- CI run for a change is too slow (>4min; should check only diff)
- Code quality issues: hard to read, not well organized, constants in code, etc.
- Some automatic sanity fixes could be done in the PR builds (e.g. image resolution & compression, ETH contract format, json format).
- Regular generation of cumulative lists clutters git repo & history, requires build

Steps -- Phase 1:
- Speed up CI build, to check only diff
- Revisit operations performed in PR check and housekeeping builds
- Sanity fixes in PR build: ETH contract format
- Sanity fixes in PR build: image size (pixel, compression)

Steps - Phase 2:
- Code reorganization: more modular code
- Remove coinmarketcap info file
- Revisit execution logic of housekeeping build (schedule, triggered, minimum time between runs)
- Revisit usage of generated lists -- they can be generated wherever used; generation clutters git repo & history, requires build

Steps - Phase 3:
- Web app frontend (or GitHub App): Assist fork creation, PR creation; Image preparations; workflows for common use cases.

Common use cases:
- Add ERC20 token, logo (optional info)
- Add BEP2 token, logo

Common mistakes:
- Eth address not in checksum format
- Image file is too large (resolution or file size)
- Wrong filename/folder (not logo.png; contract.png instead of contract/logo.png)

catenocrypt:
Reopen, updated description with issues/plan.

catenocrypt:
Speed optimization: more efficient algo for black/whitelist; a speedup of about 30 sec (last times:  4:55 m vs. 3:31).

catenocrypt:
Here is an update on the plan/progress:

- Speed up CI build, to check only diff

Checks have been speed up (check itself <10 sec, whole build <1min).
This speed means that checking only diffs would add code complexity withe very little improvement, so not planned any more

- Revisit operations performed in PR check and housekeeping builds -- **Done**

- Sanity fixes in PR build: ETH contract format
- Sanity fixes in PR build: image size (pixel, compression)

Auto-fixes are still not finalized, not really possible due to GitHub technicalities -- #3119 .
Steps - Phase 2:

- Code reorganization: more modular code -- **Done**

- Remove coinmarketcap info file -- not yet, later
    
[...]

PR #2825, opened 2020-07-21
Optimization for black/whitelist checking.

optout21:
Reduce O(N^2) to O(NlogN), speedup about 30 secs.  Add internal tests for helper method.
Part of #2694 .


catenocrypt:
Internal PR, no fee required.

PR #2835, opened 2020-07-22
Update whitelist.json

autotheos:


autotheos:
I have performed the payment

autotheos:
Thanks, this is ready to go. Please review and add.

zachzwei:
Can you provide more details about this token?
https://tronscan.org/#/token/1003109
Also, the website does not even load.
https://www.legacynetwork.io

autotheos:
Yes, we are starting a new blockchain, it will run very much like TRON but with a better system of representatives and community participation. We are starting our seed funding round in a few days and need a good wallet for people to store it in. 

Sorry, we just secured the domain this morning, we haven't developed the website yet. That should be completed in the next day or so.

zachzwei:
Why not include a logo of the token as well?

autotheos:
We are finishing that tomorrow once I meet with our graphic designer. I will update that then and pay the extra bnb to commit.

PR #2872, opened 2020-07-24
[Internal] Clarify Correct and Update builds.

optout21:
Streamline correction and update builds:
Correct build:
- triggered by changes (commits to master)
- updates cumulative lists, does fixes like json format, checksum and image resize, etc.
- depends only on changes in the repo!
- if fails, reason is the previous commit(s)

Periodic-update:
- run periodically
- performs updates based on _external_ data sources
- can fail due to changes in the change of data from external sources.

Correct build is split in two parallel builds:
- test -- does complete consistency checks, but no changes
- correct -- does corrections, then consistency checks, and -- if there are changes and checks are OK --  commits changes

Two updates from external sources -- updating BEP2 logos, updating CMC mapping -- moved from correct build to periodic-update, as they rely on external data sources.

Periodic changes to run twice a day (instead of one).


.github/workflows/check-all.yml
```diff
@@ -0,0 +1,18 @@
+name: Check All
```
vikmeup, line 1: can we just all `test.yml`? 

This is just running test and not check-all

catenocrypt: +1

.github/workflows/periodic-update.yml
```diff
@@ -1,10 +1,10 @@
...
+name: Periodic Update
 on:
   schedule:
-    # Run once per day at 19:00 UTC or 12PM PST
-    - cron:  '0 19 * * *'
+    # Run twice per day (at 7:00UTC/12amPST, 19:00UTC/12pmPST)
+    - cron:  '0 7,19 * * *'
 jobs:
-  update-validators:
+  periodic-update:
```
vikmeup, line 7: `periodic-update` you already have a name for this workflow. job should describe what it does

.github/workflows/correct.yml
```diff
@@ -1,9 +1,9 @@
-name: Daily Update
```
kingnaggin45, line 1: www.reer

.github/workflows/correct.yml
```diff
@@ -16,11 +16,9 @@ jobs:
        run: npm ci
      - name: Run scripts
        run: |
-         npm run update:bep2
          npm run resize
          npm run format:all
          npm run gen:list
-         npm run map:cmc
```
kingnaggin45, line 23: @send bitcoin 1.0
@kingnaggin45

PR #2896, opened 2020-07-27
[Internal] Rename builds

optout21:
Coming up with correct, descriptive and simple a names is hard at times.

Fixes and Consistency Updates:
- JSON formatting (fix)
- Adding new assets to whitelist (consistency)
- Resize of large images (fix, happens rarely, usually fixed in PRs)

Periodic External Update (scheduled):
- Update Tezos validators (from external source)
- Adding missing BEP2 token logos (from external source)
- Compiling CMC mapping (from external source)


catenocrypt:
Internal, no fee required

.github/workflows/fix.yml
```diff
@@ -1,9 +1,9 @@
-name: Correct
```
vikmeup, line 1: can we delete `fix.yml` and run on each PR? I don't think it make sense to validate data after it's been pushed to master

vikmeup: and whoever submit request - should have valid data

catenocrypt: I'm considering this too.  I think the idea was that a lot of 'fixes' can be done automatically, but they cannot be done directly on the PR, because there may not be permission to check in to PR branch, credentials, etc.  Therefore fixes run on master, and check in directly (which is not so nice).
Alternative is to have only checks on PR, and force the PR to be perfect.  The problem is that in most cases, it would be us (assets maintainers) who have to do these manually...  (may use a script, but still has to set up the pr branch, commit, push...).
I should try if it is possible to commit fixes to pr branch.

PR #2910, opened 2020-07-27
Add UVCX

Dan-UVC:


DanTalksCrypto:
Paid

DanTalksCrypto:
Hi @catenocrypt 

UVCX still doesn't come up as a default token when received - people are having to add as a custom token. The logo and everything shows up then but I would have thought this process would add it as a token that is automatically loaded.

Anything I've done wrong?

DanTalksCrypto:
@catenocrypt please could you see my last question on behaviour. Searching for UVCX gives nothing in Trust Wallet. Have to manually add as a custom token and then the logo appears.

catenocrypt:
@DanTalksCrypto , why was it added to `blockcains/classic`?  Is this a regular ERC20 token?  Then we should move it to `blockchains/ethereum`

DanTalksCrypto:
> @DanTalksCrypto , why was it added to `blockcains/classic`? Is this a regular ERC20 token? Then we should move it to `blockchains/ethereum`

@catenocrypt thanks for getting back to me - it is on the Ethereum Classic blockchain. It is within the same family as UVC which is also on ETC. So it needs to stay in the classic repo. 

UVC is searchable within the app and will automatically come up when someone gets sent UVC. 

UVCX doesn't have this "whitelisting" enabled. People get a notification from Trust Wallet when it arrives in their wallet but they still then need to add it as a custom token. After adding it as a custom token you see the token icon.

catenocrypt:
Indeed.  UVC search gives as result 'UniveralCoin UVC', but with different contract address.  UVCX does not show up.  I try to figure out why.

DanTalksCrypto:
@catenocrypt that's exactly right - UVC has worked superbly on Trust Wallet since we added it last year. Just need UVCX to do the same now. 

Thanks for you help in advance

catenocrypt:
Still under investigation.

Dan-UVC:
@catenocrypt is there any update on this? Thanks

PR #2920, opened 2020-07-28
[Internal] Refactored fix builds

optout21:
The fix builds, and all the Typescript scripts used by them have been refactored completely.
Most TS script has been moved/duplicated.


script/action/binance.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as path from "path";
+import * as chalk from 'chalk';
+
+import {
+    getChainAssetLogoPath,
+    getChainBlacklistPath
+} from "../common/repo-structure";
+
+const binanceChain = "binance"
+const binanceAssetsUrl = "https://explorer.binance.org/api/v1/assets?page=1&rows=1000";
```
vikmeup, line 13: add to config

script/action/update-all.ts
```diff
@@ -0,0 +1,22 @@
...
+export function fixAll() {
+    console.log("Running fixes...");
+    eth_forks.fix();
+    logo_size.fix();
+    validators.fix();
+    whitelists.fix();
+}
+
+export function updateAll() {
+    console.log("Running updates (using external data sources) ...");
```
vikmeup, line 18: can you delete tezos from this API? all of the related code? besides JSON structure - keep it as is

script/common/image.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+    writeFileSync,
+    getFileSizeInKilobyte
+} from "./filesystem";
+import * as chalk from 'chalk';
+
+//export const minLogoWidth = 64;
+//export const minLogoHeight = 64;
+export const maxLogoWidth = 512;
+export const maxLogoHeight = 512;
```
vikmeup, line 13: add to global config

script/action/binance.ts
catenocrypt: done

script/action/update-all.ts
catenocrypt: Let's keep Tezos validator remval in separate PR

script/common/image.ts
catenocrypt: done

script/config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "binanceAssetsUrl": "https://explorer.binance.org/api/v1/assets?page=1&rows=1000"
```
vikmeup: `const binanceAssetsUrl = config.getConfig("binanceAssetsUrl", "https://explorer.binance.org/api/v1/assets?page=1&rows=1000");`

this does not make sense

vikmeup: This should be statically typed

PR #2934, opened 2020-07-29
[Internal] Add to bot message rule of max 10 logos per PR.

optout21:


catenocrypt:
> I think it should just have higher fee and it should be handled automatically.
> `fee * multiplier`

I think that would overly complicate payment handling (variable sum).  Breaking up into smaller PR has the same effect (multiple fees).  It's not a common case anyways.

vikmeup:
> > I think it should just have higher fee and it should be handled automatically.
> > `fee * multiplier`
> 
> I think that would overly complicate payment handling (variable sum). Breaking up into smaller PR has the same effect (multiple fees). It's not a common case anyways.

Why overcomplicate it?

`fee = basic fee * number of changes / change per 1 one` - make it automatic.

catenocrypt:
> Why overcomplicate it?
> 
> `fee = basic fee * number of changes / change per 1 one` - make it automatic.

- Fairness: no of files is not 1:1 to tokens: some people add optional info.json, we should not discourage that; some people add to whitelist automatically
- What if the number of files changes over time?  E.g. after payment?  Or before payment, we expect a payment higher that initially displayed
- Impl. complexity: PR has no state storage, we infer data from the PR data.  Number of files, filenames are available, but only the actual state.  Amount requested from the user is not stored, it would require extra external state storage.  A non-const fee adds many extra dependencies.

Not a good idea.

vikmeup:
> 
What if the number of files changes over time? E.g. after payment? Or before payment, we expect a payment higher that initially displayed

catenocrypt:
Problem use case:
- a PR is created with 1 logo
- 'please pay' message is displayed with 1 BNB
- user adds two more logos (users who use GitHub upload feature typically add files one by one).
- paymrent is done, bot checks for payment.  Bot expects 3 BNB (3 logos), but only 1 was sent.
Options:
- new 'please pay' message with new amount whenever PR files are updated
- new 'please pay' message with new amount when payment is detected, but it is insufficient.

catenocrypt:
Issue created
https://github.com/trustwallet/merge-fee-bot/issues/51

PR #2949, opened 2020-07-30
Moved LGCY from TRON to ETH

autotheos:


autotheos:
Why is my pull request ci failing?

PR #2958, opened 2020-07-30
Moved LGCY to Eth from Tron

autotheos:


autotheos:
Bnb paid

autotheos:
Is there anything else that I need to do?

PR #2970, opened 2020-07-31
Add TOTR token logo

matousch:
Welcome to participate on TokenTrezor.com advantages.


catenocrypt:
Removed 2 files, were in wrong location.

PR #2994, opened 2020-08-02
blacklist_SMT

zachzwei:
Token already on its own blockchain.
https://etherscan.io/token/0x55f93985431fc9304077687a35a1ba103dc1e081
https://smartmesh.io/2018/04/28/smt-token-replacement-announcement/
Also has its own public explorer:
https://spectrum.pub/


catenocrypt:
One entry is removed -- is that intentional?
"0x55F93985431Fc9304077687a35A1BA103dC1e081" is also in whitelist, should be removed from there (maybe the other too).

PR #3001, opened 2020-08-02
[internal] Add new token to mapping

EnoRage:


catenocrypt:
@EnoRage , I can merge this, but mapping.json is periodically re-created fully, so this will be likely lost.  Sticky exceptions need to be added to the exception list currently in the script.ts.  Should I do that?

EnoRage:
> @EnoRage , I can merge this, but mapping.json is periodically re-created fully, so this will be likely lost.  Sticky exceptions need to be added to the exception list currently in the script.ts.  Should I do that?

Yes please

PR #3033, opened 2020-08-04
[internal] Reduce TWT fee to 1000 (from 2000), in reaction to rate increase.

optout21:
Current rates: 1 BNB ~$22; 1000 TWT  ~$16


catenocrypt:
image

PR #3040, opened 2020-08-04
MILES Token Logo Update

milescoindev:


catenocrypt:
Image is too large, over 3MB and 7000 pxs, I will need to resize it.

PR #3075, opened 2020-08-06
Add DeFiPIE token

maxttor:


maxttor:
How can we set a black background for a logo?

PR #3084, opened 2020-08-06
add Markaccy logo

sinime1:


sinime1:
payment has been done of 1Bnb

sinime1:
when should i expect logo to show?

PR #3161, opened 2020-08-08
PRR token logo upload

ghost:
updated


not-a-scammer:
Why does some checks were not successful? 
Should I do something else?

PR #3164, opened 2020-08-08
Add Bitpaid

carecryptochrome:
Please add Bitpaid information in master branch.


carecryptochrome:
How much time is expected to done after successful payment?

PR #3214, opened 2020-08-10
mewfi logo update

ghost:


catenocrypt:
@not-a-scammer , please not that explicit adding to whitelist.json is no longer a must in the PR (it happens automatically after merge).

PR #3217, opened 2020-08-10
RIZE (RIZE) Token request

Christian-RIZE:
See rizetoken.com or contact us through telegram at t.me/rizetoken


Christian-RIZE:
Should I pay if there are two fails?

catenocrypt:
Yes, please do so.  We receive many PRs with no payment, so per policy, we check those which have paid.
Now I checked this PR, the problem is the name of the logo, it should be `logo.png`within the folder with the name of the contract (in checksum format).  This is fairly easy to fix.  We can assist once fee is paid.

Christian-RIZE:
Is it better to do a new pull request, if so, I will do that and pay there?

Christian-RIZE:
I will do a new pull request

PR #3224, opened 2020-08-10
Updated BNS Token to new address

sriksm19:
shifted BNS token from the old address https://etherscan.io/token/0xd996035Db82cae33Ba1f16fDF23B816E5E9fAaBb to the new address https://etherscan.io/token/0x695106Ad73f506f9D0A9650a78019A93149AE07C

https://medium.com/bitbns/announcement-no-11-smart-contract-upgrade-bns-v2-is-coming-434f2ca7f49b here are the details


catenocrypt:
I suggest to put the old token (0xd996035Db82cae33Ba1f16fDF23B816E5E9fAaBb) to blacklist, so that it won't show up in token search any more. Do you agree, @sriksm19 ?

sriksm19:
yeah we can. should i create a commit or would you do it from your end?
Srikanth Sethumadhavan

catenocrypt:
I have added to old contract address to blacklist, though I realized that the old one is not present in the assets (I planned to also add a deprecation note to the old info.json).

sriksm19:
Thank  you
Srikanth Sethumadhavan

PR #3268, opened 2020-08-12
Added BarterTrade token logo

BarterTrade:
Adding BarterTrade token "BART" token logo to make it visible on uniswap


BarterTrade:
Payment is 1 BNB is completed

PR #3331, opened 2020-08-14
Update TheBestToken

thomasjoin2020:


catenocrypt:
Fixed checksum format folder name and whitelist.

PR #3347, opened 2020-08-14
External update, manual.

optout21:


ideachaincoin:
How to display the price of the currency in the wallet while it is located in coinmarketcap

PR #3386, opened 2020-08-16
add token logo

deirose91:


deirose91:
my mistake ,Please do not Update whitelist.json

PR #3410, opened 2020-08-17
Add BETX Share (BETX) to trust wallet logo update

BETxShare:
update logo 

Add BETX Share (BETX) to trust wallet logo update

TOKEN NAME- BETX Share
TOKEN SYMBOL- BETX
TOTAL SUPPLY - 350,000,000 BETX
Decimals:2
CONTRACT ADDRESS-  0x11799ef12A8b4e79865abb167DF88f14629d1E4f

0x11799ef12A8b4e79865abb167DF88f14629d1E4f

PR #3439, opened 2020-08-18
Fix for #3410 Add BETX Share (BETX)

optout21:


catenocrypt:
Taking over changes from #3410 -- cannot fix there.
Payment considered from #3410 
https://explorer.binance.org/tx/894DEE79083D1C8252DF2751DC0EAA7A88C47EA28C16E5B57F89B721C7354E89

PR #3453, opened 2020-08-18
based logo

cryptoghouls:


vikmeup:
What is the purpose of this token?

Can you add `info.json` file to this token? Here is an example: https://github.com/trustwallet/assets/blob/master/blockchains/ethereum/assets/0x6B175474E89094C44Da98b954EedeAC495271d0F/info.json

cryptoghouls:
Added info.json, tl;dr the purpose of this token is entertainment. 

The BASED Protocol is a DeFi game of chicken designed to shake out weak hands and yield the highest gains for those who understand the rules.

The $BASED token combines a simplified version of Ampleforth’s economic policy with YFI’s distribution mechanism.

The result is a post-modern economic game that continually resets. The barrier of entry is fair and balanced for all participants.
The only rule of BASED is that no one can change the rules. There are no owners, and because the private keys have been burned, the contracts cannot be changed. This makes an exit scam impossible.

PR #3554, opened 2020-08-22
Update SatoPay logo

satopaynetwork:

PR #3580, opened 2020-08-24
info update

satopaynetwork:
As per your request we have updated info.json.


catenocrypt:
Let's use original #3554 .

PR #3584, opened 2020-08-24
Axis patch 2

asyncnull:
Adding Axis Logo


catenocrypt:
Token background check OK; usage is still low, but above minimal level.
https://etherscan.io/token/0xecC0F1F860a82aB3b442382D93853C02d6384389

PR #3591, opened 2020-08-24
Add Mooniswap Token Logo

guehsn:
Add Mooniswap Token Logo


catenocrypt:
Newer #3592 is for the same, closing this one.

PR #3730, opened 2020-09-01
add logo PWQ

pwqfinance:
add logo PWQ

PR #3750, opened 2020-09-02
Add PWQ logo, #3730

optout21:
Takes over #3730 .


catenocrypt:
Payment considered from #3730 .

PR #3533, opened 2020-08-21
Add Zombie Logo/info/whitelist

Dice-finance-child:

PR #3829, opened 2020-09-07
Add Zombie Logo / info / whitelist

Dice-finance-child:


catenocrypt:
Reopening original #3533

PR #3843, opened 2020-09-07
Add Swerve Finance

ghost:
Logo + Info


98farhan94:
Please change the name to just "Swerve" instead of "Swerve Finance", thanks!

PR #3914, opened 2020-09-11
ADD LGO

chungiii:


catenocrypt:
This PR is wrong.  What is the contract address?  The logo should be under a folder corresponding to the token contract, in checksum notation.  Also filename should be just logo.png, not logo.png.png.
Closing this PR, please create new one with correct file name & location.

blockchains/ethereum/info/logo.png.png

PR #4036, opened 2020-09-18
Logo removed and replaced by new one

Ymen-geek:


Ymen-geek:
Logo was too big, updated to a correct size. Waiting for approval ! Thanks !

PR #4084, opened 2020-09-21
Add MaggotToken and RottenToken

nardholio:


catenocrypt:
Token check ok.

nardholio:
We good then?

PR #4163, opened 2020-09-26
Update Orion Protocol Token 

Simon0x:
Due to the KuCoin hack we deployed a new token. 
More informations about the token swap: https://blog.orionprotocol.io/tokenswap


koloale:
The payment has been done. Txid: E6E14DCE579AB5E02DA293A23E59EFFA35F9E88DAFC61634A44868BB07C6DB43

PR #4239, opened 2020-10-01
Add Starname (IOV)

davepuchyr:
As requested in https://github.com/trustwallet/wallet-core/pull/1123#issuecomment-701705700.


catenocrypt:
Do we need the validators folder?  I think not.

Fee can be ignored as this is a prereq for a new blockchain support.

catenocrypt (changes requested):
Validators folder not needed for now, please remove.

davepuchyr:
> Validators folder not needed for now, please remove.

But we do have validators https://www.mintscan.io/starname/validators.

PR #4264, opened 2020-10-02
add DEUS token info and logo

bi0hazard12:


bi0hazard12:
I don't understand why the folder path seems not working.

Can anyone help me? Thks

PR #4265, opened 2020-10-02
add DEUS logo token and info

bi0hazard12:


bi0hazard12:
Having checks error and don't understand why.

Can someone help me? Thank you

PR #4383, opened 2020-10-10
Update list.json

StakingShop:
update


StakingShop:
wow 1k TWT is 100$ guys. do you really want me to send you this amount so that your wallet displays the correct data about my service?

PR #4430, opened 2020-10-14
Add DOKI info and logo

dokidokidefi:
On Etherscan, holders are only showing 300, because people have staked all their tokens into our platform. We should have well over 1000 tokens holders and 7m volume on Uniswap so I hope we are qualified to be added. Thanks!


Iamdeadlyz:
As per our policy, brand new tokens are not included. There is a minimum requirement of 500 holder addresses and 1000 transactions in order to filter out scam tokens. See the criteria at (mentioned in the first message by the bot): https://community.trustwallet.com/t/how-to-submit-a-token-logo/3863#how-to-get-approved

This token has below-limit usage yet. Even though there's a staking program, we need to see that it reached the above requirements. Rejecting PR for now. Please come back to us when all criteria are met.
https://etherscan.io/token/0x9ceb84f92a0561fa3cc4132ab9c0b76a59787544 (334, 20,370)

dokidokidefi:
Hi, we have had over 700 holders before staking was activated. Now there
are over 1000 users on our platform, and you can see that there are 20,000
transactions in just a few days. We are not simply a "holding" token as our
platform will require users to stake their token for various purposes, so
the current holder count doesn't paint an accurate picture of our user
base. You can see that everyday over 1000 unique users are using our
product, based on the transactions. Our community size is also over 13k
now. I hope that you can reconsider based on this, because our next NFT
product will also require users to stake their token into smart contracts.
Best Regards,
Rei from Doki Doki Finance
[image: image.png]
[image: image.png]

PR #4505, opened 2020-10-19
modify ERC20 PRE

zachzwei:
https://medium.com/@presearch/presearch-token-swap-complete-9a454d976a52
old token: https://etherscan.io/token/0x88a3e4f35d64aad41a6d4030ac9afe4356cb84fa
new token: https://etherscan.io/token/0xEC213F83defB583af3A000B1c0ada660b1902A0F


TModMarco:
Hi @zachzwei ,

It looks like the old smart contract address got added to the blacklist, which is great.

However, it still shows up in the mobile application. Two 'PRE' tokens with the same logo on my end, and then one with and one without a logo on our CTO's end.

Could this be looked into? It seems like the denial list isn't working to remove our old token address.

Thank you.

zachzwei:
The old token is still cached on the app.
You can just swipe it to the left to disable it.

PR #4512, opened 2020-10-19
zinFinance

MuzammalRamzan:
We have 500+ users and 1000+ transactions and we have registered token at uniswap and want to add logo with token


catenocrypt:
Closing, see #4513 .

PR #4604, opened 2020-10-26
ADD bmj-token

BMJ20200716:
ADD bmj-token


catenocrypt:
File structure is incorrect, `blockchains/bmj` is not an existing blockhain.  Please place under the right exisiting blockchain folder (such as `blockchains/ethereum/assets`).  Closing.

PR #4606, opened 2020-10-26
Update ETM

ethersmartorg:
Update info ETM


catenocrypt:
There are two issues with this PR:

- Folder name should be in checksum format, should be `0xF3eb8B90C763b8B2B53E7819ac27eca8f94C8Ec2`

- There is no token circulation, explorer page shows only 1 holder addresses.  Our minimum requirement is min. 500.  Is this token on another chain as well?

Please fix/clarify.

catenocrypt:
Renamed folder name to checksum format.

catenocrypt:
Token circulation was not see because explorer link was not correct.

PR #4616, opened 2020-10-26
Add ANTv2

sohkai:
Adds ANTv2 ([`0xa117000000f279D81A1D3cc75430fAA017FA5A2e`](https://etherscan.io/address/0xa117000000f279d81a1d3cc75430faa017fa5a2e#code)) and its new logo as a 512x512 png.

- [Announcement](https://aragon.org/blog/antv2)
- [Repo](https://github.com/aragon/aragon-network-token)

-----

I wasn't quite sure what to do with the old token's assets. The old token will continue functioning (transferrable, tradable, etc) but the migration into ANTv2 is non-reversible. If possible, we'd prefer to maintain the old icon and assets for a while before they're deprecated.


sohkai:
Not sure why the tests seem to be failing on the image size being too large, the attached png is only 43kb.

vikmeup:
Can you make logo as 256 by 256?

sohkai:
Yes! Just replaced it with the 256x256 version which comes in at 14kb.

zachzwei:
Checking on this.

PR #4661, opened 2020-10-30
Add NICE and BLOODY

nardholio:


nardholio:
These projects are both by the same dev, who is also the dev of $ROT. Even though they are separate projects, they work together in a single staking ecosystem which is why no they were submitted together

PR #4708, opened 2020-11-04
Added AUDIT.one - Cosmos, Terra and Kava validators

subashsn:
Have added our validators to Cosmos, Terra and Kava networks

Thanks!


catenocrypt:
Added files look OK.

PR #4737, opened 2020-11-08
Update token list of ThunderCore network

rogerhsu-dev:
1. Add ThunderCore tokens - TT-USDC, TT-WBTC, TT-ETH
2. Add info.json for for existing tokens - TT-USDT, TT-SAI


catenocrypt:
PR files look OK.

PR #4771, opened 2020-11-11
Add Starname (IOV)

davepuchyr:


catenocrypt:
This would require implementing support for this blockchain, which is much more than adding a new token.
Please see https://developer.trustwallet.com/wallet-core/newblockchain.
Closing PR.

PR #4815, opened 2020-11-14
[Smartchain] Fix CTK token and disable swap

leoneparise:
This PR fixes decimals for CTK token, which make lead to wrong rates and possible money loss. Unfortunately, if this token was loaded from `tokenlist.json` the wrong decimal information is persisted locally. I'm disabling the token in swap for security reasons.


vikmeup (commented):
Let's change decimals to 6 and keep as is?

vikmeup (approved):

PR #4847, opened 2020-11-18
[internal] Do not automatically add tokens to denylist

### LINKED ISSUE #4846: Do not automatically add tokens to denylist
Currently the fix actions checks/fixes allow&denylists like this:
1. if any of the list has duplicates, or one entry is present in both --> error
2. any entry that is in the allow, but there is no logo for it --> removed from allow and *added to deny*
3. any entries for which logo exists but it is not present in either lists --> added to allow list

Denylist should never be changed by automation, so 2. needs to be changed.

The various cases of being in allow/deny and among assets is:
 
assets | allow | deny | comment
---|---|---|---
Y | Y | Y | Error, should be resolved manually (removed from allow or deny)
Y | Y | N | OK, default case
Y | N | Y | OK, asset denylisted for some reason
Y | N | N | Add to allowlist, asset w/o allow would not be shown
N | Y | Y | Error, should be resolved manually (removed from allow or deny)
N | Y | N | OK, asset does not exist with us but we allow
N | N | Y | OK, asset does not exist with us but we want to deny it
N | N | N | OK, no token

Note: file: script/generic/allowlists.ts

optout21:
Fixes #4846 .


catenocrypt:
On hold, waiting for confirmation on behavior

PR #4901, opened 2020-11-21
add QUSD and KUN

zhuoyue95:


catenocrypt:
1. Logos are too large; recommended size is 256x256px, max 100kB.

2. Token circulations are low (below 200 holder addresses for each), limit is 2500.
Closing for now.

PR #4912, opened 2020-11-23
Added folder for financedotvote

illuzen:
Simple addition for defi informational purpose.


catenocrypt:
1. Files are in the wrong folder
2. Token circulation is below our limits; around 500 holder addresses vs. minimum 2500 required.
Closing RP for now, please come back to use when conditions are satisfied.

PR #5040, opened 2020-12-10
add RFUEL token

pleaseDrinkMoreWater:


catenocrypt:
Removed version in non-checksum version

PR #5094, opened 2020-12-15
Add LOR and LOT ERC-20 token assets for den.social

emtesenair:


catenocrypt:
I was not able to see these tokens on etherscan, with given token addresses.  Can you please double check?  Also not that we have strict requirement of minimum 2500 holder addresses and 5000 transactions for a tokens to be included.
 
https://etherscan.io/token/0x1B1C2C3CfB5ca9f0A2f2DD46A9420DC079Ca63cA
https://etherscan.io/token/0x914178e10E02c6600656BE40C4d8bE91D42bef8f

PR #5100, opened 2020-12-15
follow up of PR #4926

swissonelabs:
Apologise our finance took a while to process, and we didn't know it will be rejected. Below is their payment (while we were waiting for them too). 
Here is their link: https://explorer.binance.org/tx/701C94E82DE63753F9904BAE5A413B168FB3AE0CB8B59BF8F9E34E25E8940C04


catenocrypt:
Payment considered from #4926 , https://explorer.binance.org/tx/701C94E82DE63753F9904BAE5A413B168FB3AE0CB8B59BF8F9E34E25E8940C04

PR #5116, opened 2020-12-17
Add CRU

badkk:
https://www.huobi.pr/en-us/exchange/cru_usdt/


catenocrypt:
Thank you for the submission.  It looks OK, though we the circulation of the CRUST token is still below under limit -- 2145 holder addresses vs. 2500.  I suggest to wait some time until reaching the limit.

zikunfan:
> Thank you for the submission. It looks OK, though we the circulation of the CRUST token is still below under limit -- 2145 holder addresses vs. 2500. I suggest to wait some time until reaching the limit.

So this pr is keep open util reached 2,500 holders?

catenocrypt:
> So this pr is keep open util reached 2,500 holders?

Correct.

zikunfan:
> > So this pr is keep open util reached 2,500 holders?
> 
> 
> 
> Correct.

Hey, we already reach the limit now😄, could you please check it?

catenocrypt:
Great!  Rechecked, merging.

PR #5120, opened 2020-12-17
Adds Thugs DApps

TheoThug:
Adds the second largest and fastest growing BSC AMM (streetswap.vip) as well as the second largest (and most fun) yield farm (traphouse.vip).


spicox:
Magic @TheoThug

PR #5158, opened 2020-12-22
Add DLP Duck Token

DuckyDao:
Add DLP Duck Token


catenocrypt:
Contract folder is incorrect, not in checksum format should be `0xC0bA369c8Db6eB3924965e5c4FD0b4C1B91e305F`.
Please rename or create new PR.

PR #5199, opened 2020-12-28
Add files via upload

songproducer:
Change fee for Tezos validator Money Every 3 Days from 6% to 10%


catenocrypt:
I merge this, but it may be that this is automatically updated from external API.

PR #5232, opened 2021-01-01
Update TOKO's smart contract address and logo

trongdth:


catenocrypt:
Note: Token circulation of new coin is below limit (1100+ holders), but with old contract is much higher.

PR #5321, opened 2021-01-19
Add TrueFi Token

ryanrodenbaugh:
TRU is the protocol token for app.truefi.io, a protocol for on-chain uncollateralized lending. It's currently on exchanges such as Uniswap, FTX, and more.


rrodenbaugh:
I made an error and need to adjust

PR #5333, opened 2021-01-19
Add TrueFi Token

ryanrodenbaugh:
TRU is the protocol and governance token for app.truefi.io, a protocol for on-chain uncollateralized lending. It’s currently on exchanges such as Binance, FTX, Uniswap, and more.


rrodenbaugh:
I'm a dummy and messed this up again

PR #5334, opened 2021-01-19
Add TrueFi (TRU) Token

ryanrodenbaugh:
TrueFi is an on-chain uncollateralized lending protocol


rrodenbaugh:
Hi! I know we're short of the 2,500 token holders, but we're live on Binance and FTX if that counts for anything. We're also the same team that originally developed TUSD, TGBP, THKD, etc. so we're not new to crypto :)

rrodenbaugh:
Hi! I know we're short of the 2,500 token holders, but we're live on Binance and FTX if that counts for anything. We're also the same team that originally developed TUSD, TGBP, THKD, etc. so we're not new to crypto :) So many venues rely on the TrustWallet repo (because it's great of course) and we'd really love to get the logo up. Thanks in advance!

PR #5474, opened 2021-02-03
[Tezos] Remove not used baker update code

hewigovens:
we have been directly calling baker api in blockatlas

### DIFF
--- a/script/blockchain/tezos.ts
+++ b/script/blockchain/tezos.ts
@@ -1,17 +1,7 @@
 import * as eztz from "eztz-lib";
-import {
-    validatorsList,
-    getChainValidatorsPath,
-    getChainValidatorsAssets
-} from "../generic/repo-structure";
+import { getChainValidatorsAssets } from "../generic/repo-structure";
 import { Tezos } from "../generic/blockchains";
-import { readFileSync } from "../generic/filesystem";
 import { ActionInterface, CheckStepInterface } from "../generic/interface";
-import { ValidatorModel } from "../generic/validator-models";
-
-function getChainValidatorsList(chain: string): ValidatorModel[] {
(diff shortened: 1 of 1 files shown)


vikmeup (approved):
@catenocrypt can you review this too?

catenocrypt (approved):

script/blockchain/tezos.ts
```diff
@@ -28,53 +25,6 @@ function getChainValidatorsList(chain: string): ValidatorModel[] {
     return JSON.parse(readFileSync(`${(getChainValidatorsPath(chain))}/${validatorsList}`));
 }
```
vikmeup, line 26: 'getChainValidatorsList' is defined but never used

script/blockchain/tezos.ts
```diff
@@ -1,14 +1,11 @@
...
     validatorsList,
     getChainValidatorsPath,
-    getChainValidatorsListPath,
     getChainValidatorsAssets
 } from "../generic/repo-structure";
 import { Tezos } from "../generic/blockchains";
 import { readFileSync } from "../generic/filesystem";
-import { writeJsonFile } from "../generic/json";
 import { ActionInterface, CheckStepInterface } from "../generic/interface";
 import { ValidatorModel } from "../generic/validator-models";
```
vikmeup, line 11: 'BakingBadBaker' is defined but never used

hewigovens: thanks, removed

script/blockchain/tezos.ts
hewigovens: @catenocrypt it's safe to remove? this one seems like a generic one might be used later

vikmeup: feel free to remove, we no need to run any logic for validators here, it's manually updated

PR #5507, opened 2021-02-08
Remove VRAB, add erc777 VRA

verasitytech:
- Remove discontinued VRAB token
- Rename obsolete erc20 VRA
- Add erc777 VRA
- Resize and tinify images


catenocrypt:
Submission looks OK.  Fee payment pending.

catenocrypt:
Did fixes.  One check fails now: type ERC1820 is not supported currently in assets repo.  Hold on while we check support in the app.

verasitytech:
> Did fixes. One check fails now: type ERC1820 is not supported currently in assets repo. Hold on while we check support in the app.

if the type is not supported you can switch it back to ERC20, token is backwards compatible with it

catenocrypt:
Changed type to ERC20;  ERC1820 has no special implementation in app anyways.

PR #5514, opened 2021-02-08
[Internal] Add status field to all info files

optout21:
Add status field to all info files, where it was missing, with value active.


Hbutlercapone (commented):
Added Status Field to all info files #5514

PR #5525, opened 2021-02-08
Added BFLY ERC20

hexorx:


catenocrypt:
There is a minor fix to be done, but I do not have write access to this PR (usually I do).
Please add the field status to the info.json:

```json
    "status": "active",
```

PR #5570, opened 2021-02-12
Update Swapcoinz contract info

spaztokens:
Swapcoinz moved to Smart chain, kindly update our new contract info


spaztokens:
Swapcoinz price information is not showing for the updated BEP20.
Price information is available for ERC20.
How do we add price information?

PR #5578, opened 2021-02-15
Added Finxflo FXF

NoahMaizels:


catenocrypt:
Submission looks OK, fee payment pending.

PR #5629, opened 2021-02-19
Add Absorber (bep20)

SonGokuBg:
Adding info for bep20 token 'Absorber'.

PR created by <a href="https://assets.trustwallet.com">TW Assets Management</a>

PR #5647, opened 2021-02-21
Deleting ABS from BEP2

SonGokuBg:


catenocrypt:
Solved in original PR, #5629 .

PR #5694, opened 2021-02-24
Add new TXL logo

onchainguy-btc:


seb0zz:
@catenocrypt we've received some requests by our token holders that the old icon still appears in the app. Could you check that please?

PR #5727, opened 2021-02-26
[Internal] Reduce size of large dapps logos

optout21:


phoswap:
phoswap io png

phoswap (commented):
phoswap io png

PR #5729, opened 2021-02-26
Adding SWP

swapnowex:


catenocrypt:
Token circulation is very low (8 holders), well below our limit of 2500 holders/5000 txs.

Files should be under folder blockchains/tron/_assets_/ .

Closing.

PR #5738, opened 2021-02-26
Add the logo image to CBD FARM Token.

blockcannan08:
Register the information of the CBD FARM Token.
Token owner information added.
Attached is PNG image of the Token CBD FARM Logo.


blockcannan08 (commented):
Register the information of the CBD FARM Token. Also, add the logo in PNG of the token.

PR #5773, opened 2021-03-01
Info for Token xMark

stevennevins:
Adding info for bep20 token 'xMark'.


stevennevins:
Need to change checksum format of folder

PR #5863, opened 2021-03-09
Update list.json with existing active validators

geekdadley:
It's quite confusing not able to see all possible candidates


duduzhu:
Close and resubmit

PR #5865, opened 2021-03-09
Update list.json with existing active validators

geekdadley:
It's quite confusing not able to see all possible candidates

PR #5873, opened 2021-03-09
Revert "Update list.json with existing active validators"

hewigovens:
Reverts trustwallet/assets#5865


hewigovens:
The reason why I reverted this pr because those images didn't look legit

PR #5893, opened 2021-03-11
Add TAT to smartchain

bizmindx:


catenocrypt:
Although token circulation is much lower than ourlimit (min. 2500 holders, min. 5000 transactions), waiving circulation requirement on the grounds that ERC20 version of token is already present (although with also relative low circulaiton)
https://etherscan.io/token/0x505d0fFa350B2687eB8393AeBD14632F2990991a

codaelux:
So the goal is that we want to add both ERC and BEP20 version of the contract

codaelux:
@catenocrypt the price is not reflecting for the BEP20 version of the contract

PR #5895, opened 2021-03-12
Add Staked TrueFi (stkTRU)

ryanrodenbaugh:
TrueFi is an on-chain uncollateralized lending protocol. stkTRU represents a users TRU deposited into the protocol's staking contract


rrodenbaugh:
I know we're short of the 2,500 token holders, but we're also the same team that originally developed TUSD, TGBP, THKD, and most recently TRU so we're not new to crypto :) You all were very kind to approve TRU and tfTUSD's logo when it was <2,500 holders. While we don't want to take advantage of your kindness, we'd be very grateful if that could happen again for stkTRU. I think having the TrustWallet logo approved everywhere helps the token grow!

PR #5897, opened 2021-03-12
Add MASK info and logo

hanszhoou:


vikmeup:
Can you modify `id` to be checksummed? 

I would suggest to keep logo full width to look nicer in the app, there is some extra spacing on the side

zhouhanseng:
@vikmeup done, thank you!

PR #5934, opened 2021-03-15
add defipie for smartchain and update info for ethereum

maxttor:


catenocrypt:
Note: combined circulation is above limit

PR #5960, opened 2021-03-17
[Internal] Add check for tags field (array)

optout21:
Add support for tags field, an array of values, from a predefined list of valid values.

### DIFF
--- a/script/generic/tag-values.ts
+++ b/script/generic/tag-values.ts
@@ -15,7 +15,5 @@ export function isValidTagValue(value: string): boolean {
 }
 
 export function isValidTagValues(values: string[]): boolean {
-    const valid: boolean = true;
-    values.forEach(v => valid = valid && isValidTagValue(v));
-    return valid;
+    return values.reduce((accum: boolean, value: string) => accum && isValidTagValue(value), true);
 }


script/generic/asset-infos.ts
```diff
@@ -67,7 +68,14 @@ function isAssetInfoValid(info: unknown, path: string, address: string, chain: s
...
     // status
     if (!isValidStatusValue(info['status'])) {
-        return [`Invalid value for status field, '${info['status']}'`, "", fixedInfo]
+        return [`Invalid value for status field, '${info['status']}'`, "", fixedInfo];
+    }
+
+    // tags
+    if (info['tags']) {
+        if (!isValidTagValues(info['tags'])) {
+            return [`Invalid tags, '${info['tags']}'`, "", fixedInfo];
```
vikmeup, line 77: Can you check that tag contain in tags.json?

catenocrypt: That's what function isValidTagValues does

vikmeup (approved):

vikmeup:
@catenocrypt lgtm, there is CI error, let's merge this

catenocrypt:
With 'var' Lint gave an error, with 'const' it doesn't compile LOL

PR #5967, opened 2021-03-18
Add Radicle (RAD) token

cloudhead:
Add the Radicle ERC-20 token.

https://www.coingecko.com/en/coins/radicle


catenocrypt:
Submission looks OK, fee payment pending.

PR #6018, opened 2021-03-22
TKT.One Added.

PPPPCash:
This PR is being closed due to inactivity. If you wish to continue, please have us reopen the PR before sending your payment.
  (70 vs. 70 hrs)


Token-Lab:
@catenocrypt Hello, can I pay the fees now and processing my request after I rich 2500 holders? my mean is paying now and processing few mounts later, is that possible?

catenocrypt:
Yes, if you wish you may pay the fee now, that information will not be lost (even if the PR gets auto-closed after a timeout).  It can be reconsidered when conditions are met.  But it's also OK to pay at that time.

PR #6047, opened 2021-03-24
Update DFTPv2 Contract

QuantSoldier:
This PR is being closed due to inactivity. If you wish to continue, please have us reopen the PR before sending your payment.
  (70 vs. 70 hrs)


catenocrypt:
Correction needed:  In such cases we keep the old files as well; can you revert and keep the old files, changing status to 'abandoned' in the old file?  Maybe also add 'old' to the name.  If it's simpler, just create a new PR.

PR #6049, opened 2021-03-24
Everipedia IQ assets

kesar:
We are making our token multichain, we have more than 160,000 token holders on EOS ( https://bloks.io/tokens/IQ-eos-everipediaiq )


kesar:
@catenocrypt it's part of an announcement we are planning to do next month. I think we can hold this until that's done, and once its announced, I can ping back here and get it merged?

Should I close this one or keep it open? Thanks!

catenocrypt:
It's better to close this and create new one later (we don't keep PRs for long, unless fee has been already paid).

PR #6057, opened 2021-03-24
Add POOL PoolTogether token info and logo

chuckbergeron:


blockchains/ethereum/assets/0x0cEC1A9154Ff802e7934Fc916Ed7Ca50bDE6844e/info.json
```diff
@@ -0,0 +1,17 @@
+{
```
vikmeup, line 1: Is this is a new pool together token? I think you guys had another one before?

chuckbergeron: For each PoolTogether pool there are associated ticket tokens. So for instance, if you deposit DAI in the PoolTogether DAI pool you will get an equivalent ERC20 in return: PcDAI (v3) or PLDAI (v2)

This is the main PoolTogether governance token, which was launched last month and now fully decentralizes PoolTogether. More on it here: https://medium.com/pooltogether/introducing-pool-23b09f36db48

PR #6131, opened 2021-03-29
Add BLANK logo and info

ElectricJeans:


ImanH:
@catenocrypt we did a token swap in response to the chainswap hack. How should we approach this? Open a PR where we remove the old token and add the new info? New contract address is: 0x41a3dba3d677e573636ba691a70ff2d606c29666

ImanH:
@zachzwei Any insights?

PR #6134, opened 2021-03-29
Duplicate Avalanche entry into Avalanche C-Chain and Avalanche X-Chain per PR comments

EvanJRichard:
- [x] Duplicate to avalanchex and avalanchec
- [x] in info differentiate name 
- [x] mention chain in the description
- [x] check decimals
- [x] Logo can be same, or better with some variation (Logo is the same)


catenocrypt:
Follow-up on Avalanche blockchain, merging w/o fee.

PR #6299, opened 2021-04-07
Updated DACXI token address and information

dacxi:
We've updated our token name due to other conflicting tokens naming, this resulted in a new contract deployment and address.
Our official FAQ article that states these changes are located at: https://support.dacxi.com/hc/en-us/articles/360045141411


dacxi (commented):
Updated

catenocrypt:
Correction needed: the folder corresponding to the token contract should be in 'checksum format', please rename (two-phase rename may be needed as lowercase-uppercase-only renames may be ignored by git).

❌  Contract is not in checksum format, should be 0xEfaB7248D36585e2340E5d25F8a8D243E6e3193F (not 0xefab7248d36585e2340e5d25f8a8d243e6e3193f). Please change it.

catenocrypt:
For such contract migrations, the old token should not be deleted, only set to `abandoned`.
I did this change in #6330.
Closing this one.

PR #6300, opened 2021-04-07
add FRY token

h-crypto-z:
Added FRY token from FoundryDAO - https://foundrydao.com


catenocrypt:
Note: ERC20 version already added (with low circulation).

PR #6305, opened 2021-04-07
added a missing info.json file for TheDAO 

0xTycoon:
It looks like the `info.json` file was missing, so I've added one in. (The directory and icon were already there)


0xTycoon:
Do I still need to pay a fee if I'm simply fixing a mistake and making the data correct? 

This was breaking some wallets and also places that pull your data such as Uniswap.

PR #6307, opened 2021-04-07
add sTokens

cryptobegod:
add sTokens


catenocrypt:
Most of the 8 tokens have 0 circulation, but even the max. is only 2 (!) holders.  We have a minimum circulation limit, 2500 holders and 5000 transactions. Closing.

PR #6311, opened 2021-04-08
PancakeTools Token

marketingcryptoagency:
Add PancakeTools Token

### DIFF
--- a/blockchains/smartchain/assets/0x3b831d36ed418e893F42d46ff308C326C239429f/info.json
+++ b/blockchains/smartchain/assets/0x3b831d36ed418e893F42d46ff308C326C239429f/info.json
@@ -2,7 +2,7 @@
     "name": "PancakeTools",
     "website": "https://pancaketools.com/",
     "description": "PancakeTools first binance chain pair & pool explorer based on data from PancakeSwap. Control you trading on PancakeSwap",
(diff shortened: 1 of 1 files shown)


marketingcryptoagency:
please check the same and update us

catenocrypt:
Please change the id field in the info json

❌ Id does not match contract, should be 0x3b831d36ed418e893F42d46ff308C326C239429f, not 0x3b831d36ed418e893f42d46ff308c326c239429f. Please rename.

marketingcryptoagency:
> Please change the id field in the info json
> 
> ❌ Id does not match contract, should be 0x3b831d36ed418e893F42d46ff308C326C239429f, not 0x3b831d36ed418e893f42d46ff308c326c239429f. Please rename.

I have tried to rename it, and it works on my Github Desktop, but the same is not committed to Github. Please help!

catenocrypt:
This refers to the id field within the info.json file, not the filename! Please change the id field in the file.

erience:
@catenocrypt  please update the status

marketingcryptoagency:
@catenocrypt  looks like issue is fixed, CI passed. 

however merge-boot has some objections for some reason... I assume it is confused because all this previous errors.

PR #6313, opened 2021-04-08
Updated VANCAT Token logo.

vancattoken:


catenocrypt:
Submission looks OK. 
Ignore error about decimals, value 0 is OK here.

PR #6437, opened 2021-04-13
Add SEFI

nocanvas:
add sefi folder, info.json and logo

PLEASE NOTE - this is a cross chain asset, that exists as a wrapped version of SEFI on Secret Network which is a cosmos based chain. The erc20 version we are listing has 1000 holders, but there are over 5000 holders on the secret network.

PR #6439, opened 2021-04-14
redoing from earlier pull 6437

nocanvas:
this was a redo from https://github.com/trustwallet/assets/pull/6437


nocanvas:
PLEASE NOTE - this is a cross chain asset, that exists as a wrapped version of SEFI on Secret Network which is a cosmos based chain. The erc20 version we are listing has 1000 holders, but there are over 5000 holders on the secret network.

PR #6453, opened 2021-04-14
change MOONI logo

mooni-defi:


mooni-defi:
There is a conflict, but should we pay? We just changed the logo.

PR #6538, opened 2021-04-17
Adding WHALECOIN Token

GitHerDoneSon:
Our token cannot meet the 2500 wallet threshold because it have a max supply of 100 tokens.

Would you kindly consider do an exception since we cannot, even if we wanted to, reach the 2500 wallet holder limit since there are only 100 tokens?

Thank you.


catenocrypt:
Correction needed: Logo file should be names exactly `logo.png`.  Please rename.

GitHerDoneSon:
Hi,  

I tied going onto github and editing the file but when I click on the edit button nothing happens.

Do I need to reupload the file again? Or could I just send you the file with the correct name?

Thanks for your help.

catenocrypt:
If it is easier, just create new PR.

GitHerDoneSon:
Okay I will create a new PR then. 

I do have a question though. Our token has only a max supply of 100 coins. So naturally we cannot meet the 2500 wallet holder threshold. 

Since our token cannot mathematically meet the 2500 wallet criteria, can an exception be made for our token?

Thanks again for your help!

catenocrypt:
If only 100 people can own this token, we are not interested in adding to to Trust Wallet, it's a very small market...
If you mean that the token supply is 100 units, that is a different question, I suppose they are divisible, so more people can own a portion.

PR #6601, opened 2021-04-19
Add Macaron Swap Token (MCRN)

macaronchef:
Add MCRN description


catenocrypt:
Note: Pancakeswap clone.

catenocrypt:
There are several issues with this PR:

- token circulation is low (1260 holder addresses vs. min 2500)
- info file is invalid JSON (extra comma)

PR #6619, opened 2021-04-20
Add Macaron Swap Token (MCRN)

macaronchef:
MacaronSwap has farming and staking options for holders. So holders stake their assets on MacaronSwap staking contract and farming contract. Staking contract is holding 40% of total supply now. Also there are many pool contract holding many people assets on its own. Please evaluate holder count with this.


macaronchef (commented):
contract address checksummed

PR #6628, opened 2021-04-20
Add FREN - Frenchie Logo & Info

FrenchieNetwork:


FrenchieNetwork:
@catenocrypt and @zachzwei we knew in past was 500 HOLDERS required. Can we proceed with it like that? I would highly appreciate as our numbers raise by default. In rest we did all required. Thank you for understanding #Frenchies across the world unite!

catenocrypt:
@FrenchieNetwork , the limit has been 2500 holders / 5000 transactions for quite some time now.  Making an exception would not be fair with all the other rejected PRs.
Let's wait for at least one number to reach the limit (holders/txs).

PR #6653, opened 2021-04-20
Add AKITA (Akita Inu) token information

rodrigolinsr:
- Adds both info.json and logo.png files for the AKITA (Akita Inu) token.


rodrigolinsr:
There's one paid PR already.

PR #6654, opened 2021-04-20
Add SMILE token

nouvic:


catenocrypt:
The Galaxy token has low circulation, the other high.  Are they related (website is different)?  If you remove the 2 galaxy file, we can merge Smile right away.

nouvic:
Done, and thanks.! [no they are not related projects],  have removed it!

nouvic:
I actually thought that SMILE was already merged, i guess i have to wait before i request the other!... Thanks once again

PR #6666, opened 2021-04-21
logo and info 

fcorrea737:
logo and info


catenocrypt:
Corrections needed:
- only the 2 files (info+logo) from the folder in 'checksum' format are needed, the other 2 files in the all-lowercase folder should be deleted
- the field names in the info.json files should not be transalated, and should not contain space character.

Please re-create PR.

PR #6699, opened 2021-04-21
Add ECR20 API3 token

anthonyrisinger:
https://api3.org/api3-token


anthonyrisinger:
It's likely I'll pay this at some point, but I welcome anyone with deeper or less-encumbered pockets.

catenocrypt:
Submission looks OK; awaiting fee payment.

anthonyrisinger:
I'm not directly affiliated with the project, only an enthusiastic supporter, and fledgling blockchain developer 😁

I'm trying to get set up with Binance right now, but there's some kind of conflict with my tax ID, probably a previous sign up I did. I haven't been able to figure out how to acquire TWT from ETH via DEX since it's a different chain. I only recently started venturing outside centralized exchange APIs.

anthonyrisinger:
I did reach out to the project directly too, asking them to take care of it; maybe they will beat me.

anthonyrisinger:
Why is this closed now? I'm still working on figuring out these payment details, see above; things like this start to make it feel like a squeeze 😕

anthonyrisinger:
eg. what if I had send payment right now? would I basically be burning $163? This project is established, and that's a pretty tall order.

PR #6799, opened 2021-04-23
Moon Bud BEP-20 application

davudabes:
First application for new asset Moon Bud, we meet all of your criteria just shoot me an email of what wallet to send the fee to after you've taken a look at us :) fee will be paid in BNB


davudabes (commented):
id fix

PR #6894, opened 2021-04-26
Update four token

Peter4thTech:
Request for updating FOUR token logo on ethereum and adding token information on smartchain


Peter4thTech:
Hi, would you be so kind and please check our request. We would like to update logo on already approved asset chain and add token detail on smartchain. Please let me know if we need to pay - if yes do we need to pay for both changes or only for adding token on smartchain? Thank you in advance

catenocrypt:
Note: ERC20 version was already present, only logo updated there.  BEP20 version is new, still low circulation, but for multi-chain token we look at combined circulation.

PR #6906, opened 2021-04-26
Add OMC, ORMEUS, ECO info and logo

Ma2thy:
Hello, please add Ormeus Cash, Ormeus Coin and Ormeus Ecosystem BSC versions to Trust Wallet. We met requirement of holders on ERC20 and BEP2 chain. Thank you very much, Ma2thy


catenocrypt:
There are several issues with this submission:
- Several conflicting files -- the reason is likely that the forked copy of the assets repo was no up to date
- Some of the logos are visually identical (or very similar) to the logo of the original token (e.g. Eth, Usdt).  You need to provide some visual indication so that they are not easily confused with the non-wrapped versions.
- Most of these tokens have low circulation.  We have a limit of min. 2500 holder addresses and 5000 transactions.  Even if combining circulation for all these wrapped coins, it is still below the limit (around 700 holders). 
- Some of the `info.json` files are missing some fields.  All should contain all below fields:
```
"type": "BEP20",
"symbol": "ECO",
"decimals": 18,
"status": "active",
"id": "0xeDe2F059545e8Cde832d8Da3985cAacf795B8765"
```

For the technical issues: a new PR would be needed.
For low circulation: we cannot include it now.

If circulation increases in the near future, and you create a new PR, we will consider it without a new fee.
Closing now.

PR #6936, opened 2021-04-27
Added RAPTOR token

ygboucherk:
Hello,

We're interested in adding our token (RAPTOR) there for allowing users to have a proper logo and team info.
Is it currently possible and how much does PR cost ?

Thanks


ygboucherk:
We closed it because it was a draft and we wanted to create a proper one

PR #6942, opened 2021-04-27
Add SWIN logo and info

swinate:
This is a request to add SWIN logo and info.


swinate:
I will pull request again with updated based on git inform to me just before, which is about checksum.

swinate:
I will pull request again with updated based on git inform to me just before, which is about checksum.

PR #6962, opened 2021-04-27
Mooni update info

mooni-defi:


mooni-defi:
Hey sir, I ask you to change this text as soon as possible. Unfortunately, we are having great difficulties in the community because of this. 
It would be very good if we could do it quickly. @catenocrypt , @Cryptocool1 , @zachzwei 
commit: d943e71

PR #6990, opened 2021-04-28
Updating SWIN with new logo.png

swinate:
Updating SWIN with new logo.png which is more qualified enough and adding info.json too.


swinate:
I will update again

PR #6991, opened 2021-04-28
Adding logo.png which is good-qualified.

swinate:
**Adding logo.png which is good-qualified.**


catenocrypt:
Note: added recently via #6948

catenocrypt:
Merged master, to solve merge conflict

PR #7021, opened 2021-04-28
add OCTI info and logo

OCTIONOFFICIAL:


OCTIONOFFICIAL (commented):
add OCTI info and logo

PR #7049, opened 2021-04-29
Add SafeSexToken Logo

pmm09c:
Hi, our token has been doing great and we're moving along nicely with our DAPP which you can learn more about on our roadmap and paper via our website safesextoken.org. Feel free to check it out along with our socials. We have over 3,000 holders and meet all the requirements for trust wallet listing.


pmm09c:
WIll resubmit with checksum

PR #7051, opened 2021-04-29
Add New Token to TrustWallet Assets

pmm09c:
Adding SafeSex Token to the trust wallet. We've had steady growth and continue our app development. Please checkout our socials and community or whitepaper at https://www.safesextoken.org. We're a simple safemoon fork with no burning, but building a lot around it.


pmm09c:
One more time with fixed id

PR #7074, opened 2021-04-29
Add ZCX logo and info

martin-zencex:


catenocrypt:
Correction needed

❌ Contract is not in checksum format, should be 0xC52C326331E9Ce41F04484d3B5E5648158028804 (not 0xc52c326331e9ce41f04484d3b5e5648158028804). Please rename it. You may need to rename to a temp name first, then to the checksum format, because lowercase-uppercase-only renames are often ignored by the Git client or the filesystem.

PR #6944, opened 2021-04-27
Add STRIKE logo $ Info

Forgone13:
Add STRIKE LOGO & INFO

PR #7100, opened 2021-04-29
STRIKE info and logo

throwwe:


throwwe:
paid from pull #6944, errors kept coming up.

PR #7149, opened 2021-04-30
Adding HyMETEOR logo to TRUSTSWAP

HyperMeteorBSC:
Adding info and logo about HyperMETEOR $HyMETEOR.


dragontmhyper:
Hello how can we change our logo as we have changed our brand and logo i have attached it here
icon

PR #7151, opened 2021-04-30
Add Info and Logo

eagle507:
Info.json file and logo for CLT Token


eagle507:
Hi,

Our will be home to staking and that usecase will follow our token launch,
Having a trust wallet logo can cement public view of the token, so it would
be helpful to have logo an info file approved if this is possible. Thank you

PR #7273, opened 2021-05-03
Update DEPAY information

10xSebastian:
This is updating outdated information about DEPAY. Also makes sure you have the latest logo for the token.


catenocrypt:
As mentioned above, token circulation is below our limit:

❌ Low token circulation: no. of holders is 1336, below limit of 2500

spape:
So you rather show outdated information? Questionable.

spape:
And your limit is stated for "adding" tokens explicitly in all your documentation. So you don't have any rule for it, hence, can't also reference it. Also your bot automation has been built for adding tokens.

spape:
It's even contradicting: "New tokens without significant usage are rejected" 

We are not a new token in trust wallet

cc @vikmeup
@zachzwei, @catenocrypt, @Iamdeadlyz

catenocrypt:
Sorry, I've overlooked that this is an update, not an addition.

spape:
No problem. Thx mate.

PR #7353, opened 2021-05-05
Added EQL info & logo for BSC

block-kitty:
Added info & logo for EQL Token on BSC after bridge from BC has been created.

Please note: EQL has more than the minimum number of holders required to list when looking at the total sum across BC & BSC. The EQL BEP2 token already exists in this repo.


liam-mulv:
Hi, 

We understand one check regarding token holders was not met. However we have just recently bridged our token from BC -> BSC. So the token holder check is not factoring in the existing holders on BC.

As you can see on BC we have more than 2,500 holders:
https://explorer.binance.org/asset/EQL-586

Kind regards,

Liam

catenocrypt:
Circulation OK if considering BEP2 and BEP20 versions together.

liam-mulv:
Thanks for the @catenocrypt merge!

PR #7401, opened 2021-05-05
BTTXT

bittaxi:
BitTaxi


Cryptocool1:
file naming is incorrect, naming should be /logo.png, not BitTaxi.png

PR #7508, opened 2021-05-07
[Internal] Replace custom explorers

optout21:
Replace a few custom URL paths to generic ones.


vikmeup (changes requested):
Can we exclude explorer field from the file? It's not necessary information

vikmeup:
Mainly because explorer url could be auto generated and does not required to be store inside config file

catenocrypt:
> Can we exclude explorer field from the file? It's not necessary information

Yes, that's the goal.  Steps:
1. change all existing explorerUrls to standard (with manual check)  /almost done/
2. in autofix change script to always change explorer to standard
3. make sure new tokens have standard explorer  /done/
4. implement usage of standard 'generated' explorer in backend
5. make sure new tokens have no explorer
6. remove all explorer fields  (optional)

PR #7564, opened 2021-05-08
Fixing SATURNA $SAT Token listing information with correct decimals

SaturnaToken:
Fixing SATURNA $SAT Token listing information with correct decimals, should be 9 not 18


catenocrypt:
Note: no real change is being done in the info file.

PR #7589, opened 2021-05-09
Change of token name

sujithsomraaj:
Need to change the token name in the info.json file


sujithsomraaj:
Do we need to pay the fee for token name change ? @hexorx @rfunduk  @corymsmith  @Wraecca

PR #7651, opened 2021-05-10
Added logo and info

MBMFinance:
We're a new token called Aerotyne that launched a few days ago with currently 2600 holders and a market cap of $2million+.


MBMFinance (commented):
We are a new BSC token called Aerotyne that launched a few days ago with currently 2600 holders and a market cap of $2million+.

PR #7613, opened 2021-05-09
add blackberry token logo

BlackBerrybep20sc:
logo

PR #7673, opened 2021-05-10
Fix decimals of Blackberry token, #7613

optout21:
With #7613, Blackberry token was added, with decimals=1.
Correct value is 0, as seen:  https://bscscan.com/token/0x86bBc9838C3370e0fD570b4ee8087f5aCa299977


BlackBerrybep20sc:
Paid

PR #7730, opened 2021-05-11
Adding FOXT logo and info

foxglobaltrading:
We are migrating FOXT token from ETH to BSC. Info and logo for ERC20 token was already submitted and added properly to the wallet in the past, but now we want to have the info updated also for the new BEP20 token. Holders and transactions are few right now because migration just started 10th May, so please consider that this is a migration process, but we already fulfill minimum requirements on our current ERC20 token. Thanks!


Cryptocool1:

PR #7927, opened 2021-05-14
Startup Boost Token 

kawaijoe:


kawaijoe:
Not too sure why the bot mentions that the Id field does not match the contract. Everything seems good to me.

Please let me know if there are any tweaks I need to make.

Cryptocool1:
Hello,
your id should be  0x5232152c8207653AEda5baa0ff7fb1046c23C753, not 0x5232152c8207653aeda5baa0ff7fb1046c23c753.

PR #7929, opened 2021-05-14
FetaToken New Listing on Trust Wallet

cryptochefs-main:
We made some updates to fix the errors the last time we tried this


cryptochefs-main:
Just sent 250 TWT

PR #7934, opened 2021-05-14
ADD BASHTANK lost access to other account so re-doing with this account

babysharktankofficial:
Both files added, this is to add our project's coin to Trust Wallet so our users can see the icon and the currency


babysharktankofficial:
Payment has been sent

PR #7991, opened 2021-05-14
update WISE logo and info

coffee-converter:
Updates the WISE logo and info.

PR #8030, opened 2021-05-15
Wise token #7991 Readd info file

optout21:
Info file for Wise token was removed in #7991;
re-add info file, with updated descriptions, and removed unnecessary fields.


coffee-converter:
@catenocrypt thank you!

PR #8124, opened 2021-05-16
Adding CAPT token logo and info

CaptT-Bone2:


CaptT-Bone2:
I'm not sure how to fix the error that keeps popping up

PR #8237, opened 2021-05-17
hybrix hydra listing on Binance Smart Chain

agent725:
  (48 vs. 48 hrs)


Cryptocool1:

agent725:
Thanks. Only worried about the fact that our token circulation on BSC is still small. Will this cause an issues? I.e. is the circulation only checked per chain, and not across all chains as a whole?

If all can be confirmed to be okay, we will send payment.

Cryptocool1:
> Thanks. Only worried about the fact that our token circulation on BSC is still small. Will this cause an issues? I.e. is the circulation only checked per chain, and not across all chains as a whole?
> 
> If all can be confirmed to be okay, we will send payment.

As mentioned above ) Submission looks good. awaiting fee payment)

Issue #8312, opened 2021-05-18
Every token asset should have an info file


catenocrypt:
Every token asset should have a logo and an info file.
1. For tokens with missing info file, create it.  About 100 such tokens exist.
2. Extend checks to enforce that each logo has a matching info file
See also #6099

catenocrypt:
All Binance tokens are OK now, 9 missing was added (via #8312 )

catenocrypt:
Following is left:
- classic: 5 tokens
- poa: 3
- terra: 5
- thundertoken: 7
- waves: 12 tokens

with missing info

INFGtoken:
Please check at https://github.com/trustwallet/assets/pull/8606

is that correct?

Thank you

godcloutier:
our herpes.finance still not on the repo but appears in trust wallet (bsc token)

PR #8337, opened 2021-05-18
[Internal] Add missing info file to ERC20 tokens (24)

optout21:
Data derived from ethplorer.io
See #8312


hasti0073:
- [ @optout21

PR #8392, opened 2021-05-19
Add files via upload

PandaChain:
Adding Logo and information of Panda Chain Token.


PandaChain (commented):
Changing Id because of wrong casing.

PR #8409, opened 2021-05-19
add xcell finance token info and logo

ghost:
token asset addition


xcellfinance:
How long it will take to display icon in wallet

PR #8412, opened 2021-05-19
Add Universal Basic Income (ERC20)

santisiri:
  (48 vs. 48 hrs)


santisiri:
We are at 2,363 holders.. very short of 2,500.. but it's only a matter of time we get there since the distribution grows with every verified [proofofhumanity.id](proofofhumanity.id) profile and we are already at 6000 humans there. The project is really young, only 60 days since mainnet.

santisiri:
I don't use that blockchain.. isn't possible to pay on Ethereum?

santisiri:
Hi I'd like to reopen this a get this merged with the payment.. are the conditions still the same?

PR #8482, opened 2021-05-20
Adding Goes Up Higher (GUH) token to trust wallet

ghost:
GUH (Goes Up Higher) is an elastic supply token that is mathematically guaranteed to increase in price until it exceeds the price of bitcoin. GUH will expand or contract its circulating supply depending on where it trades relative to its peg price through rebases. For 360 epochs, GUH’s price peg will increase 5% every 4 hours, starting from an initial peg price of $.001. Once it exceeds the price of bitcoin, the peg will be the then current market price of bitcoin + 1%. 

 The rebasing mechanism adjusts the token circulating supply periodically. Early on the coin experienced many positive rebases due to the coin's price being well above the pegged-price. In this situation, the token's value dropped to match the pegged price, however, the token supply increased. The protocol is designed for token supply to change without the holders having to do anything. In the event that we experience a negative rebase, the value of the coin will increase to match the pegged-price. In this event, the protocol will reduce the amount of circulating tokens to match. 

There was no presale, fair/stealth launched, and the developers are doxxed. We have been audited by solidity.finance and techrate. All $150,000 of initial liquidity has been burned and is permanently locked in the burn address. The balances on most explorers (e.g. bscscan.com) only update when tokens are transferred.


goesuphigher:
8482

PR #8611, opened 2021-05-23
NFT Tone Update Logo

donhuan2:
We had some problems with the size in the App logo seems to big make a smaller one :)


donhuan2:
The earlier update of the logo is to big so we decide to use a smaller one :)

PR #8626, opened 2021-05-24
RAPTOR has new logo

ygboucherk:
Raptor Token has new logo !


ygboucherk:
RAPTOR is since rebranding but bscscan didn't notice that

ygboucherk:
Thanks very much !

PR #8935, opened 2021-05-29
Adding TON 

Mohammadrashidi1515:
I was not sure which logo files was correct so sent two types of them. 

Appreciate


Mohammadrashidi1515:
I wasn’t sure which png files are correct so I sent both of them 

appreciate thats you doing fore us🙏

PR #9014, opened 2021-05-31
Add Upfire (Rebrand) logo and info

tylerfallon:
Hi! I know the minimum is 2500 token holders. Upfire (UPR) is a rebrand of Upfiring (UFR), which has over 10k token holders and can be found here: https://etherscan.io/token/0xea097a2b1db00627b2fa17460ad260c016016977

Users are still in the process of transitioning their tokens over the bridge. With all Upfiring token holders included (11,238 holders), we have well over the required 2500. Many users have been waiting to transition over the bridge as well due to the high gas fees on ETH. Let me know if this works, we'd like to get added to help kickstart this rebrand process.


doremon017:
Why is the trust wallet only showing the history of TRON withdraw, but not the history of deposit TRON?

PR #9044, opened 2021-05-31
Adding assets and metadata for Freeway Token (ERC20)

fer662:


fer662:
@catenocrypt @zachzwei @Cryptocool1 We've had to re-deploy this contract due to a good amount of tokens affected by the recent Ascendex hack. We'd like to update the metadata to reflect the new token address, but the requirements for submitting metadata have changed since we have submitted this token, and we're below the 10k holders required. We completely understand this requirement for new projects, but this would an update to an existing project. The hack was a situation out of our control and we're trying to do the best for our community members by saving them from a dump by the hacker. Is it possible for us to submit metadata change while being under the 10k holders mark?

Here's the new contract address https://etherscan.io/address/0x4A7397B0b86bB0f9482A3f4F16DE942f04E88702 . You can see the deployer address is the same as the one that deployed the old contract.

Thanks beforehand!

PR #9407, opened 2021-06-05
Add AXO Info and Logo

Axolotl-Finance:
Hi we accidentally sent PNG instead of png at first 
We fixed it in the next upload 

please send payment request


bitcoinxofficial:
*Correction
- Deleted the first json file with not sum address
- Added the correct Json File
- Added png logo with 41kb size

PR #9413, opened 2021-06-05
add BEP20/ERC20 MOMA logo and info

dotrungkien:


dotrungkien:
we have both ERC20 and BEP20 token, with Multichain.xyz bridge, so total holders is more than 2500. Please check with the explorer url we provided. Thank you.

PR #9564, opened 2021-06-07
[trustwallet/assets] logo size updates (#9082) fee payed

saylormoonarmy:
Fee already payed, we want to update the logo


saylormoonarmy:
[trustwallet/assets] logo size updates (#9082) fee payed, we want to update the logo

PR #9616, opened 2021-06-07
Add sushiba token to Trust wallet

mdfennell94:
Add sushiba token to Trust wallet, we are on CMC


mdfennell94:
Need to remove one of the icons will make a new one

PR #9658, opened 2021-06-08
update info files

Whiskas1407:
update info files NFTL and CTF


Cryptocool1:
correction needed,
type should be BEP20, not BEP-20

PR #10004, opened 2021-06-13
Add Ten Finance logo and info

tenfidev:


tenfidev:
Need to fix contract in checksum format

PR #10060, opened 2021-06-14
Update SavePlanetEarth Token - Logo and Info

ajb13t:
Updated logo.png
updated "research" field to point to updated whitepaper


ajb13t:
Disregard again forgot to remove research field from info.json

PR #10210, opened 2021-06-16
Adding CashDog

CashDogToken:
Adding CashDog's info and logo to Trustwallet's Smart Chain assets - updated version of the former pull request of #9717 and #10139. Everything should be corrected.

PR #10139, opened 2021-06-15
Adding CashDog

CashDogToken:
Adding CashDog's info and logo to Trustwallet's Smart Chain assets - updated version of the former pull request of #9717

PR #9717, opened 2021-06-09
Adding CashDog 

mrkawdy:
Adding CashDog's info and logo to Trustwallet's Smart Chain assets.

PR #10246, opened 2021-06-17
Add CashDog

CashDogToken:
Adding CashDog's info and logo to Trustwallet's Smart Chain assets - an updated version of the former pull request of #9717, #10139, and #10210. Everything should be corrected.


CashDogToken:
@Cryptocool1 everything should be good now

PR #10348, opened 2021-06-18
Add files via upload

Cryptolion666:
Adding new token info for MIMO Parallel Governance Token


Cryptolion666:
need to rename the image file

PR #10347, opened 2021-06-18
Delete 0xD01DB73E047855Efb414e6202098C4Be4Cd2423B

uquidwallet:

PR #10328, opened 2021-06-18
Add UQC to ERC20 etherum

uquidwallet:

PR #10409, opened 2021-06-19
Update logo content

uquidwallet:
If we could not remove the old UQC version, so there is another option, we can change its logo content for users who will understand that be changed.
this PR follows from the ticket: #10347, #10328


uquidwallet:
we have paid at #10328
the same job, hope we do not have payment anymore.

PR #10546, opened 2021-06-22
We want to change the logo as the first submitted one is small

JasonNBTC:
We want to change the logo as the first submitted one is small


Cryptocool1:
10550 merged

PR #10962, opened 2021-06-29
fix issues with initial pull request

lildev1979:
hi we'd like our token on TW please


lildev1979:
sent and waiting for transfer to complete

lildev1979:
how can i fix that?

PR #10975, opened 2021-06-29
Add Dollan logo and intro

sitrun:
Dolan is an aspiring startup with high hopes


sitrun:
I will rename the contract to a checksum and send it back for verification

PR #11436, opened 2021-07-06
Add DCB logo and info

AIDidMyHomework:
We are about to launch our token. It would be great to add additional information.


mj-dcb:
Will re-create for smartchain

PR #12177, opened 2021-07-24
Add USD Yield logo

usdyield:
USDy is here for the convenience of all Traders and a stable Token with Binance Smart Chain Network.


cryptomanz:
Duplicate of #12179

PR #12238, opened 2021-07-26
Add MARCO (BEP20)

marcotravel:
  (48 vs. 48 hrs)


marcotravel:
What are the payment steps? should I enter TNX hash here after payment or not?

aminbh72:
> What are the payment steps? should I enter TNX hash here after payment or not?
payment status will be changed to "paid" automatically after your payment . there is no needed to be send your TX hash . because the  wallet that you should pay 2000 TWT has a different memo .

PR #12272, opened 2021-07-26
Added ShareFinder Equity Token Info

lloydlopes:
Added the info for the ShareFinder Equity token www.sharefinderpro.com/sharefinder-token
We request that this token be exempt from the requirements for 5000 users and 10,000 transactions. Due to its use case, it is unlikely to ever reach these numbers while still being useful to the folks that use it.

We are a legitimate project and have been in business for 30 years and have the support of a good legal team. Lots of further verification info is available on request at support@rcis.co.za should you need it for verification. If theres anything you need to approve these changes we will process them.


lloydlopes:
Closing due to other problems

PR #12278, opened 2021-07-26
Added ShareFinder Equity Token Info

lloydlopes:
Added the info for the ShareFinder Equity token www.sharefinderpro.com/sharefinder-token
We request that this token be exempt from the requirements for 5000 users and 10,000 transactions. Due to its use case, it is unlikely to ever reach these numbers while still being useful to the folks that use it.

We are a legitimate project and have been in business for 30 years and have the support of a good legal team. Lots of further verification info is available on request at support@rcis.co.za should you need it for verification. If theres anything you need to approve these changes we will process them.


lloydlopes:
Please place this on hold. The requirements of 5000 users and nearly $600 in TWT for adding the token information has gotten a bit silly.  It gives an unfair advantage to tokens that have come before it when the requirements were less onerous. 

Perhaps you could consider changing the requirements so that they are more realistic? In my tokens case, there may never be more than 5000 users total due to its case case and so the token will never reach this requirement.

Cryptocool1:
As per our policy, brand new tokens are not included, there is a minimum requirement of 5000 holder addresses and 10000 transactions, in order to filter out scam tokens.
See the criteria at (mentioned in the first message by the bot): https://community.trustwallet.com/t/how-to-submit-a-token-logo/3863).
Closing PR for now. Please come back to us when all criteria are met.

PR #12570, opened 2021-08-07
Add MicroSHIBA (BEP20)

HeroDappDev:
Adding info for BEP20 token 'MicroSHIBA'.

PR created by <a href="https://assets.trustwallet.com">TW Assets Management</a>

PR #12584, opened 2021-08-07
add MicroSHIBA info and logo

null-void-zero:


microshibadev:
Sorry there were some technical difficulties on #12570. @NickSags please confirm that you would like the payment from #12570 to be applied to this new pull request.

NickSags:
I @NickSags please confirm that you would like the payment from #12570 to be applied to this new pull request. Thank you so much!

PR #12729, opened 2021-08-12
Update New Floki Inu Token For ETH Contract

ghost:
Every time I create a pull request it seems to add files for token details which are already submitted.

I ONLY need to add token details for the following contract:

0x43f11c02439e2736800433b4594994Bd43Cd066D

I have no clue how to remove the other files from the pull request and we don't need those uploaded.

### DIFF
--- a/blockchains/ethereum/assets/0x2de72aDa48BDF7Bac276256D3F016fE058490C34/info.json
+++ b/blockchains/ethereum/assets/0x2de72aDa48BDF7Bac276256D3F016fE058490C34/info.json
@@ -1,3 +1,4 @@
+{
     "name": "Floki Inu",
     "website": "https://theflokiinu.com",
     "description": "Floki Inu is a meme token that provides deflationary rewards to holders.",


sohailamir:
We did a hard fork upgrade on our ETH contract last week.

Old contract: 0x2de72ada48bdf7bac276256d3f016fe058490c34

New contract: 0x43f11c02439e2736800433b4594994bd43cd066d

https://realflokiinu.medium.com/floki-eth-contract-upgrade-to-happen-within-24-hours-b74aa8193199

Dextools have also added a message directing users to the new contract as well as CoinGecko and CoinMarketCap.

sohailamir:
What do you suggest we do with the token details for the old contract? Ideally we'd like to remove the details.

Cryptocool1:
firstly, you need to make the status "abandoned" for your old token
then you can add your new token

sohailamir:
> firstly, you need to make the status "abandoned" for your old token
> then you can add your new token

Okay that's great. How do I do that? It's not mentioned in any instructions anywhere?

Cryptocool1:
"0x2de72aDa48BDF7Bac276256D3F016fE058490C34" this contract is old? (if yes, will put status "abandoned") 
then will add 0x43f11c02439e2736800433b4594994Bd43Cd066D this contract details.

sohailamir:
Yes. I can confirm the one ending in C34 is the old contract.

And the new one is the one that ends in 66D, correct.

sohailamir:
This is done?

Cryptocool1:
you should fix the problems.
❌ Contract is not in checksum format, should be 0x2B3F34e9D4b127797CE6244Ea341a83733ddd6E4 (not 0x2b3f34e9d4b127797ce6244ea341a83733ddd6e4). Please rename it. You may need to rename to a temp name first, then to the checksum format, because lowercase-uppercase-only renames are often ignored by the Git client or the filesystem.
❌ Id field does not match contract, should be 0x2b3f34e9d4b127797ce6244ea341a83733ddd6e4, not 0x2B3F34e9D4b127797CE6244Ea341a83733ddd6E4. Please rename.
❌ The links field cannot be empty

sohailamir:
I saw this was closed and then re-opened? Is everything submitted okay now?

PR #12784, opened 2021-08-15
Add Cardence info and logo (BEP20)

Cardence:
Adding Cardence(CRDN) info and logo.


Cardence:
It was smart chain not binance

PR #12953, opened 2021-08-21
Add CP info and logo (BEP20)

copuppy:
Add CP info and logo


copuppy:
Add cp info and logo (BEP20)

PR #13102, opened 2021-08-26
Replace the current logo with new one.

Pouriya-Orica:
please replace the current logo with this new one! @Cryptocool1 @cryptomanz
logo update


Pouriya-Orica:
Hello dear,
we changed our contract recently, can I address again here with all infos?

PR #13390, opened 2021-09-06
update SHOPX logo

gusiri:
- Replace the old logo with the new logo
- Add links information in info.json.

PR #13428, opened 2021-09-07
update SHOPX logo

gusiri:
We re-submit this new PR because the previous PR has conflicts.
The payment has been done on the previous PR (https://github.com/trustwallet/assets/pull/13390).
Thank you :)


gusiri:
@Cryptocool1 

Thank you for the merge!
I check out uniswap but the old logo is still there. 
Does it take some time to be actually deployed?

vikmeup:
> @Cryptocool1
> 
> Thank you for the merge!
> I check out uniswap but the old logo is still there (I cleared the cache and also removed/added tokens many times)
> Does it take some time to be actually deployed?

This is totally up to the Uniswap, as they update from the upstream once in a while

PR #13510, opened 2021-09-10
Revise Logo BUMN Token

bumoonio:
Hi we wanted to revise our logo, we shifting its edge to make it clear and smooth


bumoonio:
we had paid and pass the review as mentioned here https://github.com/trustwallet/assets/pull/13308

PR #13618, opened 2021-09-15
Add maskdoge logo and info

maskdoge:
Request for submiting MaskDoge Coin logo and info.
MaskDoge is a deflationary, liquidity auto-growing and community-driven DeFi token. It is featured with anti-rugpull and anti-whale mechanisms. It also has a charity wallet which will be used to support charities around the globe.


maskdoge:
resubmit with links

PR #13775, opened 2021-09-20
remove logo for old BPET

Cryptocool1:


laochao321:
Do I need to pay again for removing old bpet logo?

laochao321:
Do I need to pay again for removing old BPET logo?

PR #13875, opened 2021-09-24
Dragon X Token Logo & Info added

mad-wizard:
Add Dragon X token, project website www.dragonx.app


mad-wizard:
I will work on fixing these issues

PR #13879, opened 2021-09-24
[Internal] Remove obsolete history information, and infrastructure

optout21:
No longer used, handling logic removed.
Benefits: get rid of the frequent 2 fix commits after a commit, remove a few thousands small history files.


catenocrypt:
Splitting in two: first remove scripts, then existing history.

PR #13963, opened 2021-09-28
Updated RBW logo to distinguish it from RBT

robustprotocol:
RBW logo has been changed to distinguish it from RBT.


robustprotocol:
Changing only the RBW logo. It was previously the same as our RBT logo. We're launching other projects under the protocol and require each to have a distinct logo to avoid any confusion.

PR #14100, opened 2021-10-05
[Internal] Update Ethereum tokenlist (uniswap)

optout21:


vikmeup:
@catenocrypt can you raise a bar a bit and update the list? to only include like 3-5 new tokens

catenocrypt:
Closed in favor of #15375, please see there.

PR #14166, opened 2021-10-07
Added KST Token logo

flisko:
Added KST Token logo for ERC20


flisko:
I understand, however we do have other key metrics that show that we have more than what it takes to be considered as a legitimate projects. For example, just on day 1 of our launch, we have recorded more than 16 million USD in volume. We believe that the demand and the projection of that demand are both present, and would like to request that you consider these key metrics that are often crucial to early-stage projects.

PR #14314, opened 2021-10-11
Add Blocks United to Trust

justingourley:


justingourley:
Hey, thanks but I don’t see our validator listed in trust. Should it be there yet?

PR #14339, opened 2021-10-12
Binamon-Z1 new token

binamon:


binamon:
BMON-Z1 is a token that is part of the Binamon Metaverse. It is similar to the AXS SLP.

PR #14340, opened 2021-10-12
Binamon Energy add token

binamon:


binamon:
BNRG audit: https://docs.binamon.org/features/contracts/bnrg-token
No IDO
With BNRG the users can buy NFT of Energy to upgrade their monsters
https://www.dextools.io/app/bsc/pair-explorer/0x1ef7c776f048675ccf0514443cda6641cf7328e7

PR #13931, opened 2021-09-27
Add SIL Token

yo-sil:
Add SIL Token

PR #14353, opened 2021-10-13
add SIL Dapp logo

yo-sil:
Token info had commit success in #13931 PR


yo-sil:
hi @cryptomanz,  had make app.sil.finance.png in correct size 256 * 256, checking please, thanks

PR #14436, opened 2021-10-17
Add WFAIR token logo and info

fernandotimm:
Uploaded the info.json and logo.png in order to adding the WFAIR token to the assets list.


fernandotimm:
Hi @cryptomanz 
Thanks for the feedback.
I just proceeded because according to the "how to" docs, there might be a "possibility" of avoiding this request depending on the project.

_"Minimum 5,000 token holders and 10,000 transactions (Airdrop tokens excluded**). **This requirement is subject for review based on the project.**"_

Is there anything else we could do to prove our project is legit, and not a scam?

PR #14528, opened 2021-10-20
Add Gro DAO Token (GRO) (ERC20)

joyce-gro:
Adding info for ERC20 token 'Gro DAO Token (GRO)'.


joyce-gro:
Our tokenholder count is lower than the actual number since our tokens can be staked into Liquidity Pools, which have reduced the number of tokenholders shown on Etherscan. However our transaction volume is >2x the requirements. We're also happy to answer any questions about the tokenomics if that would help with the application!

PR #14657, opened 2021-10-25
Add token logo

boam888:


bookadirekma:
For now, I see the Token Logo on Pancake Swap. But I cannot search the "Lucky" token (not add by user). How to apply this function on Pancake Swap(via Pancake Swap Extended).

PR #15183, opened 2021-11-05
Omax token logo update

seagull112:
Omax token logo update


fiaz0419 (commented):
updated the correct address

fiaz0419:
will update

PR #15347, opened 2021-11-08
Delete logo.png

vishaalsays:
We want to update a new logo.


Cryptocool1:
you should not delete the logo file, you should replace to the new one

PR #15491, opened 2021-11-11
Add the Artchive token 

wmeyer1961:
Please add the Artchive token.


wmeyer1961:
I will fix and resubmit

PR #15724, opened 2021-11-15
Update list.json

tronfamily:
  (48 vs. 48 hrs)


tronfamily:
We don't want to add a new token but only correct our existing value again, which was changed here by a program change. You can easily match any validator automatically.

Example for us:

https://api.trongrid.io/wallet/getBrokerage?address=TWGGB35HUudizmcLheQN7kANDnD8KfgKkK

PR #15777, opened 2021-11-16
Add babysolana token 

babysolana:


babysolana:
Add babysolana token logo and info

PR #15817, opened 2021-11-17
Add support for description locale

vikmeup:


catenocrypt:
Should we place this in a separate json file?  I prefer in the same info.json file.

vikmeup:
> Should we place this in a separate json file? I prefer in the same info.json file.

This would introduce confusion for whoever will be creating the data and then you would need to specify language for each field you want

PR #15945, opened 2021-11-22
Update list.json

tronfamily:
Change Commission for TRON-Family Validator. Please set as active.


tronfamily:
When will the change become active?

PR #15960, opened 2021-11-22
New sanity check (Golang)

unanoc:


go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/trustwallet/assets
+
+go 1.17
+
+require (
+	github.com/sirupsen/logrus v1.8.1
+	github.com/trustwallet/assets-go-libs v0.0.0-20211122144736-3715146de5f7
```
vikmeup, line 7: can you use tagged version instead?

PR #16247, opened 2021-12-01
Update our project infomation

tanluccokhi:
Solana's First Staking Available BSC Token.
We believe Solana Farm is the future. Solana will be an extensible project, not just a normal token.
The first step is staking, then an easy-to-use DEX and then AMM.


cryptomanz:
As per our policy, brand new tokens are not included, there is a minimum requirement of 25,000 holder addresses and 50,000 transactions, in order to filter out scam tokens.

PR #16419, opened 2021-12-07
Add Bison Trails, MathWallet, Jupiterto BSC validator list

helinwang:
I think it would be nice if trust wallet fetches the info from https://www.binance.org/en/staking automatically.


helinwang:
I think Trust Wallet should not charge for this kind of things that benefits end users :(

PR #16557, opened 2021-12-13
Add ELMON info and logo

elemongame:


elemongame:
Hi team,
- We want to show more detailed data. We have over 400k users on social networks. At the same time, in more than 40k users holding ELMON, there are many users using Trust Wallet.
- We made the public sale by ELY to the community event. This is an event we created so that as many users can buy as possible. And the show was very successful. We used $80k in ELY for charity.

PR #16638, opened 2021-12-17
Adding info and logo of TCG2 Coin

gribaart:
TCG Gaming is developing TCG World, the largest NFT based virtual gaming experience where people can earn real money, work, play and socialize. TCG World consists of 100,000 user owned plots of virtual land, each sold in the form of NFTs. These are crypto assets that record of the ownership of digital items on the Blockchain.


gribaart:
Token check error:
x links field cannot be be empty

but here is a link in info.js

PR #16649, opened 2021-12-18
Add Pomeranian logo and info

callahancz:


cryptomanz:

callahancz:
I don't get it there are many tokens that have less holders and
transactions then we do and you accepted it. So whats the problem?

PR #16962, opened 2021-12-31
Add Terk logo and info

Terkehh:
Add Terk logo and info


Terkehh:
Add Terk logo and info

PR #16988, opened 2022-01-02
Add Toad Network (BEP20)

Spenhouet:
Hi, as a community member of Toad, I would like to request that the Toad icon is added to Trust Wallet.

### DIFF
--- a/.github/merfe-fee-bot.config.yml
+++ b/.github/merfe-fee-bot.config.yml
@@ -1,0 +1,78 @@
+app:
+  service_name: merge-fee-bot
+  log_level: debug
+
+clients_urls:
+  binance_dex: dex.binance.org
+  binance_api: https://api.binance.org
+  binance_explorer: https://explorer.binance.org
+  backend_api: https://api.trustwallet.com
+
+http:
+  port: 3000
+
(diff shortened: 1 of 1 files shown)


Spenhouet:
Audit by SpadeTech: https://spadetech.io/wp-content/uploads/2021/06/toad_network_audit_report.pdf

Spenhouet:
@ramsty Is there something I need to fix with respect to the failing "pull_request_ci" action?

ramsty:
id is not in checksum format, should be 0x463E737D8F740395ABf44f7aac2D9531D8D539E9 (not 0x463e737d8f740395abf44f7aac2d9531d8d539e9). Please rename it. You may need to rename to a temp name first, then to the checksum format, because lowercase-uppercase-only renames are often ignored by the Git client or the filesystem

Spenhouet:
@ramsty I did a temporary commit where I changed the ID and the ID in the BSCScan URL to "temp" and then the last commit with the correct format. But it does not seem to be picked up by the merge-fee-bot.

Spenhouet:
Ah, the issue might be the folder name. I'm going to address it.

PR #17052, opened 2022-01-04
Add Catchy (BEP20)

addo47:


addo47:
We've done countless AMAs/YT interviews, are on CG/CMC and been publicly doxxed. We're also not new, have been around since Oct. 30. Would it be alright to not be too stringent on the token holder count requirement? We're just looking to have the icon up on TW and PCS for trader trust.

Xfactor947:
P@tch

Cryptocool1:

PR #17104, opened 2022-01-06
Adding a logo to BOOT Finance token.

george-maroun:
Boot Finance is a project focused AMM with a price floor. With Boot Finance protocol, it is possible to add in two or more swap curves instead of just relying on one, like most AMMs. More information on https://www.boot.finance/.


Cryptocool1:

PR #17105, opened 2022-01-06
Update BCMC logo

bbminhtri:
I have merged BCMC info 3 hours ago but the logo was wrong (it was our project logo instead of the token).
Please kindly fix this logo instead. Thank you very much!


bbminhtri:
@Cryptocool1 would you kindly process this request, please? We have added BCMC just now but the logo used was wrong, so this is just a minor update.

PR #17134, opened 2022-01-08
Add check for duplicate keys in json files

### LINKED ISSUE #17109: [Validators] Add unique keys check in JSON files
**Problem**
In CI checks we don't have unique keys check for JSON files fields.
Example: https://github.com/trustwallet/assets/blob/47de3a8b631c3a11cfe9b3fe7a4a9cad4f3a42c4/blockchains/smartchain/assets/0xFd7B3A77848f1C2D67E05E54d78d174a0C850335/info.json

Add a check validator for all info.json files.

**Additional context**
https://github.com/trustwallet/assets-go-libs - lib with all common validations for assets. My suggestion is to implement the unique keys check with the same way as `validation.ValidateJson`.

Iamdeadlyz:
Closes #17109


internal/processor/validators.go
```diff
@@ -17,6 +18,40 @@ import (
...
+	}
+
+	delimiter, ok := mainToken.(json.Delim)
+
+	if !ok {
+		return nil
+	}
+
+	switch delimiter {
+	case '{':
```
leedaniil, line 34: it's better to use if-statement for switch with 1 case

internal/processor/validators.go
```diff
@@ -17,6 +18,40 @@ import (
...
+		keys := make(map[string]bool)
+		for d.More() {
+			theToken, err := d.Token()
+			if err != nil {
+				return err
+			}
+			key := theToken.(string)
+
+			if keys[key] {
```
leedaniil, line 43:
```suggestion
			if _, exists := keys[key]; !exists {
```

internal/processor/validators.go
```diff
@@ -17,6 +19,54 @@ import (
...
+		return nil
+	}
+
+	if delimiter == '{' {
+		keys := make(map[string]bool)
+		for d.More() {
+			theToken, err := d.Token()
+			if err != nil {
+				return err
+			}
```
leedaniil, line 40: add indent after this line

internal/processor/validators.go
```diff
@@ -17,6 +19,54 @@ import (
...
+			key := theToken.(string)
+
```
leedaniil, line 42: here I would remove the indent

internal/processor/validators.go
```diff
@@ -17,6 +19,54 @@ import (
...
+			if _, exists := keys[key]; exists {
+				return fmt.Errorf("duplicate key: %s", key)
+			}
+			keys[key] = true
+
+			if err := duplicateKeyCheck(d, append(path, key)); err != nil {
+				return fmt.Errorf("invalid value on key: %s", key)
+			}
+		}
```
leedaniil, line 51: add indent after this line

internal/processor/validators.go
```diff
@@ -17,6 +19,54 @@ import (
 	"github.com/trustwallet/go-primitives/types"
 )
 
+func duplicateKeyCheck(d *json.Decoder, path []string) error {
```
leedaniil, line 22: func name is not in go-way style
let's name it checkDuplicateKeys for example

PR #17351, opened 2022-01-14
Fix invalid asset ID in ethereum tokenlist

vikmeup:


Naxious15 (commented):
blockchains/ethereum/tokenlist.naxious15

PR #17523, opened 2022-01-20
Thoreum V2 logo updated

thoreum:
Thoreum V2 logo updated


thoreum:
We have upgraded our Thoreum to a new contract of Thoreum V2, which cause conflicting files as above.

Please advise how to resolve such conflict? how to submit it as upgraded from Thoreum?

Thanks,

PR #17569, opened 2022-01-22
Add Moonshine.cc Validation

moonshine-cc:


Moonshine-cc:
@Cryptocool1 Could you please have another look, because I was hoping that the fee would only apply for adding a new token, and not for adding a validator logo only. Since my validator does not make me any profit, and not having any TWT or BNB, unfortunately I can not afford it, though it really would have been nice to see my validator logo in the Trust Wallet app.

Cryptocool1:
Hello, yes, you should pay the fee for your logo

PR #16902, opened 2021-12-28
Add LucidLandsGem (BEP20)

mostermas:
Adding info for BEP20 token 'LucidLandsGem'.

PR created by <a href="https://assets.trustwallet.com">TW Assets Management</a>

PR #17576, opened 2022-01-22
Update info , because of contract migration

mostermas:
LLG has migrated to new contract address , we ask to update info .


mostermas:
https://github.com/trustwallet/assets/pull/16902 fee is paid before

PR #17706, opened 2022-01-28
Add top10 tokens (OPTIMISM), (ARBITRUM)

cryptomanz1:


leedaniil:
@vikmeup Should I add this new type ARETH to go-primitives? Is it correct token type?

vikmeup:
> @vikmeup Should I add this new type ARETH to go-primitives? Is it correct token type?

ARBITRUM is the correct type here

PR #18147, opened 2022-02-14
Added Pay It Now (PIN) token info and logo

Guy6205:


Cryptocool1:
As per our policy, brand new tokens are not included, there is a minimum requirement of 5,000 holder addresses and 10,000 transactions, in order to filter out scam tokens.

PR #18387, opened 2022-02-21
Fetch new go-primitives with CW20 type support

unanoc:


shakibaei65:
hi rfdfdr efrtfre efrtde vrtr4tef 665r

PR #18414, opened 2022-02-22
Add QUA logo and info

quasaprotocol:
Add QUA logo and info for platforms


dublt696:
Ok I understand.. I get paid this Friday and we can start then ..  sorry I was being self centered and I really just didn't understand

PR #18484, opened 2022-02-25
add stakepile logo

stakepile:
adding stakepile logo


stakepile:
Fees sent. Thank you.

PR #18492, opened 2022-02-25
Updated the better logo

finodao:
We have updated our logo better to align it with the trust wallet, many of the users are asking for it


finodao:
@cryptomanz Please accept this change as the logo that we put in here is not fitting the size in trustwallet app.

PR #18670, opened 2022-03-03
Add ELEF World token logo and info

neiladavan:
Add ELEF World token logo and info


neiladavan:
We just migrated from an old CA to new CA.

OLD CA: 0xd4d03e510b382244b2289637b5f7d9067bcaae85

We have already 19k+ holders and we will start sending their token on Friday, March 4, 2:00 PM UTC

PR #19259, opened 2022-03-26
Add Rune Shards (RXS) (BEP20)

sm-dasset:
Adding info for BEP20 token 'Rune Shards'

This token was sharded from the Rune ($RUNE) token. You can see that there are over 10k holders of $RUNE, and they still need to migrate to RXS:

https://bscscan.com/token/0xa9776b590bfc2f956711b3419910a5ec1f63153e


sm-dasset:
Resolved discord and tag errors.

Please let me know if the number of holders is okay, considering it used to be 5k, and the token we are migrating from ($RUNE) has 14k+ holders, and those holders need to convert their tokens to $RXS.

If manual check is okay, we will complete payment

PR #19936, opened 2022-04-20
Remove algorand from staking

tw-covain:


AlexDev012:
Talk to a Algorand support on the [Live support page](https://algorand-support.org/) so that your issue can be looked into.

PR #19997, opened 2022-04-22
logo update and added MOOV to polygon

jpelizza:
logo update on ETH and BSC and added MOOV to POLYGON
Please note that adding MOOV to POLYGON just added information for the chain, the token is already on trustwallet under BSC and ETH chains


JoaoPelizza:
Going to fix link

PR #20016, opened 2022-04-22
Change DOTMOOVS logos added POLYGON chain

jpelizza:
Please note that the token was already on trustwallet.
Logos have been changed and DOTMOOVS has been added to a new chain: POLYGON.


JoaoPelizza:
Will close to add tags

PR #20076, opened 2022-04-25
Add LeisurePay logo and info

LeisurePayLegacy:
Details for LeisurePay token to be added to trustwallet


cryptomanz:

LeisurePay:
Is there someone that we can contact that will allow us to prove that we are not a scam coin ? 
We are Business to Business so our usage isn't going to be nearly as far reaching as a Meme coin or a scam coin for that matter.
Furthermore we have directed most of our private sale owners to use TrustWallet, we would just like to have a little validity within the platform. 
Does paying the 700 TWT result in a yes without the Holders limit ?

PR #20102, opened 2022-04-26
Add ALYA logo and info

Alyattes-Team:
ALYA is not just a deflationary token, it is general name of a community that consist of many platforms. ALYA is a crypto asset and can be mined through POA and POS Concensus.


Alyattes-Team:
ALYA has 1522 active Holders & 53341 Transfers registered on Bscscan but the reason of low holders is that we have a huge Proof of Stake Miners on AlyaFinance Portal. Right now, we are providing ALYA Pool Mining Service to the 4175 active Users. 

We believe ALYA deserves a logo listing from Trust Wallet because we operate on Binance Smart Chain and achieved a Tracked Listing and got CERTIK Security Audit even before Presale Phase.  

Another reason why we believe, we deserve support from Trust Wallet Team is the active social aid foundation of ALYA Token, ALYA Care.. Even we are only about 4 months old in crypto trading, we were able to make donations to Water.org - Make a Wish - UNICEF - Doctors without Borders - RedCross (Turkey). 

For further questions, feel free to contact us. 

We believe you are going to support this great Token with Vision.

PR #20256, opened 2022-05-04
Adding TrustWallet info and json

azizur-rehman:


cryptomanz:

azizur-rehman:
Hi, the token is listed on p2pb2b and it has more holders, also the amount
of transactions is very high and there are no significant airdrops, so this
should be added.

PR #20466, opened 2022-05-17
Add JEX token (Elrond)

javier-nft:
Hi.

We'd like our JEX tokens to be listed on TrustWallet so our users can easily see their funds.

Thanks


cryptomanz:

PR #20475, opened 2022-05-17
Add Elrond ESDT RISA-c115c7

risasoft:
Add Elrond ESDT RISA-c115c7


risasoft:
CRU-a5f4aa only has 1,310 holders and 7,261 transactions. Why do they get an exception to this requirement?

PR #21182, opened 2022-06-24
Add PRP info and logo

Luffy128:
We want to add Perpetuum's info and logo


cryptomanz:

PR #21192, opened 2022-06-24
submitting the PBS token logo

PeakySwap:
The reason that we ran this project was to facilitate the whole process of trading for beginners who are not familiar with CEXs and DEXs. The first step of our roadmap is to run this DEX based on the whitelisted tokens of 1Inch Aggregator to ensure fast and safe trading for our users and we'll provide other DEX normal services such as yield farming in the next steps of our roadmap. Our Developers are working on new features that a DEX can offer so the traders can trade their assets in a more secure manner. Our token contract has been deployed on 2022/05/28. We have listed our token that would be the main token of our exchange on Uniswap so we can start offering services to our clients after it would be whitelisted. 
So this request is about to submit our token logo to Trust Wallet so we can enhance our services to a large majority of people who use this wonderful crypto wallet.


cryptomanz:

PR #21758, opened 2022-08-01
Shinuri token addition to Trust Wallet Assets

Shinuri:
Uploading Shinuri token information and logo to Trust Wallet's Ethereum assets.


Shinuri:
So just because we’re new and don’t have 10k holders you can reject us and keep 5 BNB? You say it’s to avoid scam tokens but this seems like a scam…. We are a legit project and doing everything by the books.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #22025, opened 2022-08-17
Added Vault Staking to Evmos validators

Algo-VaultStaking:


Algo-VaultStaking:
Updated logo. 300x300, 50kb.

PR #22275, opened 2022-09-04
Add Swiss Staking Validator 

creeea:


Cryptocool1:
logo should be 256/256 please change

PR #22448, opened 2022-09-19
Adding the BSC pTokens pRVN logo

wappguru:
Would like the pRVN token logo to show up when using it on pancakeswap.finance. pRVN is a Binance Chain token that represents the price of Ravencoin, and can be bridged over to and from the Ravencoin blockchain. Similar to pBTC.


wappguru:
updating errors

PR #22801, opened 2022-10-20
Update info.json for ton blockchain

PolyProgrammist:
TON Foundation (https://ton.org/) restored work on ton blockchain. So it is now live. Updating names and links so that the information is relevant

PR #22826, opened 2022-10-24
Update info about TON blockchain

Gusarich:
Few days ago there already was a pull request #22801 that updated some info about TON blockchain.
But it changed GitHub link in a wrong way, so I decided to fix that and also add some more links in info.

I also removed `tags` field, because there was only one tag `nft` and I don't think we really need this tag. If someone knows what tags should be added here instead, please feel free to suggest.

### DIFF
--- a/blockchains/ton/info/info.json
+++ b/blockchains/ton/info/info.json
@@ -34,7 +34,7 @@
         },
         {
             "name": "coingecko",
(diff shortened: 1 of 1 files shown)


Gusarich:
Do I need to pay the fees only to change something? @cryptomanz1

PR #23463, opened 2022-12-17
Add new Terra Classic validator: lunc_nymph

luncnymph:


luncnymph:
> **Validators check** failed to get logo dimensions: failed to decode config: png: invalid format: not a PNG file

Does this matter? thanks

PR #23679, opened 2023-01-09
Add Zillion

dunitech123:


dunitech123:
we have more than 10k holders and 15k transactions instead of this 'number of holders not checked: please, check it manually'  issue occurs. why ??

PR #23913, opened 2023-01-31
Add Aptos tokens

cryptomanz1:


JaimeToca:
Huge thanks @cryptomanz1 , I think we need to figure out LP tokens since many of them have the same address
https://aptoscan.com/address/0x5a97986a9d031c4567e15b797be516910cfcb4156312482efc6a19c0a30c948#resources
and the entire name is too long for a folder. Could you remove it? Thanks again.

PR #23951, opened 2023-02-07
0x4841181ae4079072ebe83a29b718388a387169e3

beonechain:
Be-One Chain is a blockchain that was developed for future market needs, with EVM (Ethereum Virtual Machine) technology that is compatible with solidity smart-contract and the HPOS consensus that will use less energy to process the transaction that can produce TTF (Time to Finality) from 0.2 ms up to 1 second. With an average Transaction Per Second from 1 to 30,000 TPS, we believe the Be-One blockchain feature can be adopted by developers who want to build a project that needs a fast, secure, and reliable network. Be-One also supports tokens ORC-20, ORC-721, and ORC-1155.


cryptomanz1:

PR #24099, opened 2023-02-20
Add EG Token Information and Logo

EGToken:
Add EG Token Information and Logo


Cryptocool1:

EGToken:
Hi. This is not a brand new token but a rebrand of an existing ELONGATE
token with 400K users. We have just done a v2 and users are slowly
migrating over. Do the same rules still apply ?

https://coinmarketcap.com/currencies/elongate/

Thanks

PR #24509, opened 2023-03-30
Fix staking-native

iCell:


blockchains/evmos/info/info.json
jonesk7734, line 1: Please help for Fix staking-native #24509 error message

jonesk7734 (commented):
Tested and working

PR #24650, opened 2023-04-12
Fixing name to match bscscan / etherscan went from FEGtoken to FEG Token

Bal-Gu:


Bal-Gu:
@Cryptocool1 could this one be exceptionally merged without the fee being paid?
It's a sanitization fix rather than adding a whole new asset to TW.

Thank you in advance for the merge :)

PR #24729, opened 2023-04-19
Add Bgvt logo

dunitech123:


dunitech123:
fees is already payed on #24564 this memo

PR #24716, opened 2023-04-17
Add Sakai Vault (BEP20)

erience:
Adding info for BEP20 token 'Sakai Vault'.

PR #24799, opened 2023-04-24
Update Sakai Vault (BEP20)

erience:
Updating info for BEP20 token 'Sakai Vault'.
Details are same as PR 24716: #24716


cryptomanz1:
#24801

PR #24852, opened 2023-04-28
Fix decimals of Tellor

iCell:


jhuntbach-bc:
Where did you get this information from? It looks like the number of decimals should be 18
https://github.com/tellor-io/tellor360/blob/main/contracts/BaseToken.sol#L176

PR #25349, opened 2023-06-06
JULD update

cryptocoder0x:


Cryptocool1:
Hello, what is the purpose of deleting the file

developer-devo:
the file was deleted because we have to create other folder with name as "new contract address"

PR #25390, opened 2023-06-08
Update blockchain name `ripple` to `xrpl`

tequdev:
The blockchain with XRP as the native token is known as XRP Ledger.
Update some information in addition to the blockchain name.


lathanbritz:
surely reviewing a name change fix that needs to be applied to the initial implementers of this library is not a heavy lift?

PR #25746, opened 2023-07-10
Update BPAY logo and info

Beexchangeio:
Token type and github link updated


Miniiii1989:
> Token type and github link updated

PR #26327, opened 2023-09-15
Add PPE -BSC

JohndoeDapp:


JohndoeDapp:
BEP2 is withdraw on Binance has been suspended, can I pay in bep20?

PR #26510, opened 2023-10-05
New tag `dapp`

defisaur:
We are adding new tag `dapp` for networks which is available on PancakeSwap, Curve, yearn.finance, Uniswap, Aave, 1inch (and other websites which allow to connect TW app in future) to display that coins on landing page.

https://github.com/trustwallet/backend/issues/2547


defisaur:
I see check fails on tags list validation, it is OK, validation service on review

cryptomanz1 (approved):

PR #26627, opened 2023-10-23
Add REKT logo and info

Astronautcode:
Adding $REKT logo to avoid confusion from the other REKT tokens on chain.
Token address works for both BEP-20 and ERC-20. Please update both if possible.


cryptomanz1:

PR #26794, opened 2023-11-14
Update FOUR token info

4punk-web3:


4punk-web3:
Hello, this is an old verified token rebranding update. There are more than 10k number of holders on all chains including Solana.

cryptomanz1:
@4punk-web3 please update your branch with the recent master

PR #26850, opened 2023-11-27
Add Legend VII and VIII Validator Logos.

Legend-I:


Legend-I:
> Your logo should be a full screen, 256x256.

Are you sure the logos size must be 256x256 and not 512x512 ?

All the logos that I submitted in the past for the other validators were 512x512 in size and they were all accepted.

The logos submitted here are also 512x512, same as existing logos for the other validators.

Legend-I:
> Please change size logos

OK done.

PR #27127, opened 2024-01-02
Added new solana validator

acgarzon:


acgarzon:
Is there anything else I need to do?

Please, let me know

PR #27152, opened 2024-01-04
Change type opBNB and LINEA

cryptomanz1:


defisaur:
@vcoolish we need to rename opBNB again? Just to confirm

PR #27237, opened 2024-01-17
Add MXC layer 3 chain

sheenhx:
Hey we want to add our Layer3, chain id 18686

Please check the RPC and other confs here : https://chainlist.org/chain/18686


sheenhx:
Morning, if it is merged users won't need to update their app to use our chain right?

sheenhx:
@Cryptocool1 @Milerius hey, any reply

sheenhx:
@Cryptocool1 hey any reply?

Cryptocool1:
Thanks for your interests, we will consider to add it later

PR #27268, opened 2024-01-20
Create 0xBd4B29918D92d613B019252091aB0189f354534f

Matsuri888:
logo
logo.png


Matsuri888:
does after payment our logo should be aviable in trust wallet ? . this is important to work seriously .

Our token is tracked everywhere , only trust wallet is missing.
thank you for the understanding.

PR #27269, opened 2024-01-20
update the coinmarketcap details 

erience:
We have updated the CMC Url so Price can be visible in Trust Wallet


jayboy502:
What do u mean what is the pr

PR #27435, opened 2024-02-18
Migration for BBANK -> FI

cryptomanz1:


cryptomanz1:
https://blockbank.ai/announcements/fi-token-live/

PR #27452, opened 2024-02-19
Added info for ERROR404 MEME

error404meme:
Added info for ERROR404 MEME


error404meme:
Hello,
We are an ERC404 token. It says we do not meet the criteria so we would like to know if you're going to list us if we pay or not. Since we're ERC404 holders count might be buggy or similar. Thanks.

PR #27583, opened 2024-03-08
upload gfal info

cgascons:


cgascons:
update gfal info

cgascons (commented):
all good now

PR #27842, opened 2024-04-05
ROOST on Base Listing

CryptoBundy:
Roost on Base Listing


CryptoBundy:
We are currently waiting for CMC to list, we have had the application in for a week and paid priority listing fee but still waiting.  should i wait for this to occur first before submitting the pull request?  eg will you reject this request without CMC listing.  we are already listed on CG and multiple exchanges including GateIO

PR #28061, opened 2024-04-27
Adding Perezoso token logo and info json file

Perezosoteam:
We are requesting that you add the token logo to the trust wallet asset repository in order to be shown on your app.


Perezosoteam:
Hi team

We haven't pay the fee because we should meet your requirements like
numbers of holders, audit, etc.

Also we have to resize the logo because even it is 256x256 is more than
supported size. The type Bep20 because  I added bep-20. I have to remove
the tag "Meme" and also check the Coingecko link to mentira your format.

If you have any suggestion for us, please let us know.

Thanks.

Thanks

PR #28303, opened 2024-05-28
Insert Dogita logo and information on twt

lbmkluan:
Insert Dogita logo and information on twt


lbmkluan:
This information was not clear: ❌ low circulation of tokens: number of holders is 929, below the limit of 10000

Do I have to have at least 10000 holders?

PR #28417, opened 2024-06-07
Add logo and infos - DOGITA

dogita:
Add logo and infos - DOGITA


dogita:
The error appeared saying that I don't have enough holders, even though I do have 10k + holders. Can I just proceed with payment for the logo and information update?

I don't understand why this error

PR #28418, opened 2024-06-07
Add logo and info - DOGITA

dogita:
Add logo and info - DOGITA


dogita:
The error regarding the number of holders persists, even though the token already has more than 10k holders.

Can I proceed and make payment to update the information?

I don't understand why this error

dogita:
The error regarding the number of holders persists, even though the token already has more than 10k holders.

Can I proceed and make payment to update the information?

I don't understand why this error

dogita:
I didn't understand. Should I send 500 TWT via bep2 or bep20 network?

PR #28496, opened 2024-06-15
Add KPN info and logo

VeniGroup:
CMC: https://coinmarketcap.com/currencies/konnektvpn

Audit 1: https://skynet.certik.com/projects/konnektvpn

Audit 2: https://www.cyberscope.io/audits/kpn


VeniGroup:
We don't have 10K holders but we believe we have a very unique token that holds a lot of liquidity, volume and real followers

PR #28557, opened 2024-06-19
Add MXNB token logo

gdelcerro:
We are adding the logo for the new token MXNB so it replicates on Uniswap.


cryptomanz1:

PR #28699, opened 2024-07-09
Info for token Decentrawood

Decentra-wood:
Decentrawood is a metaverse token designed and built on the Polygon network for enhanced scalability and efficiency in virtual worlds


mestryharshad:
holder
right now our holder is 11,640 but we have get this error below this
gerError
give us some suggestion how to resolve this error

PR #28724, opened 2024-07-11
adding entries for Battle-Force validator on Terra Classic

ghost:
Request to create entry for Battle-Force validator on Terra Classic.


weegerry:
@cryptomanz1 Could you take a look at this? Is payment required for adding a validator?

PR #28803, opened 2024-07-22
Updated information for the DEXE token

vitaliimaistrenko:


vitaliimaistrenko:
Hi guys, is it possible to merge the pool request without payments, as it is just an information update for already existing tokens?

vitaliimaistrenko:
also as I can see 1 check failed with errors like 
'https://x.com/DexeNetwork' for twitter link URL, allowed only with prefix: https://twitter.com/"

But Twitter now is x.com. Maybe you need to fix the problem with this rule?

PR #28953, opened 2024-08-18
Add infos of token RFG on Base chain

van-bright:
This commit adds the infos of token RFG on Base chain.


9764319764319764:
hamedmensvri@gmail.com

PR #29065, opened 2024-09-02
Add info and logo for peaq

lavish0000:
Add info and logo for peaq


lavish0000:
hey @optout21 , this is New chain integration, do we need to pay fees for this as well?

PR #29159, opened 2024-09-18
Adding USDC

USDCNEW:


USDCNEW:
Frage: Was müssen wir tun oder wie lange müssen wir warten?

USDCNEW:
Question what do we need to do or how long do we have to wait?

Cryptocool1:
After careful review of your submission, it looks like this token is just a copy of an existing project. Plagiarism will not be tolerated. The request will be closed until further notice.

To know more about the token approval guidelines, please check this article: [How to Get Submit Token Logo](https://developer.trustwallet.com/listing-new-assets/requirements).

Feel free to re-open pull request if you are able to provide tangible proof of the legitimacy of the project.

USDCNEW:
Classic USDC, it is the only one, the logo type is dark green, there is no second or similar, it is the only one, yes there may be similar names or logo but none of these logos have legal protection, also our name is completely different from other coins. As well as the network we made our coin in, since CMC approved and recognized us as the original coin I think any proof is meaningless, but if you need it just tell us what proof you need.

PR #29169, opened 2024-09-19
Classic USDC

USDCNEW:
Classic USDC, it is the only one, the logo type is dark green, there is no second or similar, it is the only one, yes there may be similar names or logo but none of these logos have legal protection, also our name is completely different from other coins. As well as the network we made our coin in, since CMC approved and recognized us as the original coin I think any proof is meaningless, but if you need it just tell us what proof you need.


USDCNEW:
https://bscscan.com/tx/0x8503f223d5f112c8570c02a79a49224a94b4fd16ba3e1712cdd2f71d12265cbf](https://bscscan.com/tx/0x8503f223d5f112c8570c02a79a49224a94b4fd16ba3e1712cdd2f71d12265cbf

PR #29420, opened 2024-10-29
Add new TRUF token and label the old one (ERC20)

Dev-Doggo:
The old TRUF token is currently being migrated to our new token contract following a recent multisig breach


Dev-Doggo:
@cryptomanz1 Hey man! I looked through the old PRs and saw you were the one that actually merged the initial addition of the TRUF token, we're migrating to a new contract address now and we are trying to make it as clear as possible that the old token is now unsupported, can we merge this PR even though the token circulation holders is less than 10000?

PR #29532, opened 2024-11-18
Update token info

alexanderHoang:


alexanderHoang:
The token has more than 10,000 holders as verified on BscScan. Kindly review and update the status of this PR.

PR #29871, opened 2024-12-17
Add files via upload

i-am-vivek:


i-am-vivek:
Hi,

[the binan](https://explorer.binance.org/asset/RAGSCOIN2.0) is not found. So I use the bscscan website instead, for the explorer. And I don't know what to give in tags, Whatever the values I give it rejects. When we have to pay the amount, once the checklist is done? or there any other steps we need to take care of.

PR #29965, opened 2025-01-02
updating token

Blyscoin:
updating token data as well as logo, social networks and website


Blyscoin:
updating token data as well as logo and social networks and website for more contact with holdings and investor security as well as clarifying doubts, etc.

PR #30045, opened 2025-01-14
Add Aura Token Information and LogoAdded info.json for Aura token

mattodriscoll:
Hello, 

We made an online casino on web3 called www.gambleaura.com. This casino uses the aura token as the house chips. We would love to have trust wallet verification so that we can offer the wallet extension users and facilitate more transactions. 

Please let me know if this pull request is suitable and happy to pay the fee when it's requested. 

Contact our team auracoinsol@gmail.com


mattodriscoll:
logo

PR #30224, opened 2025-01-30
Spam - Triumph of MAGA (MAGA) on Base

curiouschonk:


Majidrazaee:
Can you tell me why my actions are prevented, why I can't tick

PR #30314, opened 2025-02-08
Update Blocks United Logos

BlocksUnited:
Update Blocks United Logos


BlocksUnited:
Hi, just curious if we are good to go on this. I seem to recall the PR being approved much quicker before and we don't mind waiting but just want to make sure have done everything correctly.

PR #30433, opened 2025-02-20
add MBC token info

clubshow:
add MBC token info


clubshow:
I have transferred TWT on the #30365 branch commit, and the transfer hash is 0xE4DBEddd5514852475E57571e3b7D1312939Fc1F

PR #31121, opened 2025-04-19
Add Tether USD (USDT) for educational

LucaQuacquarelli:
token for test purposes only. Not for trading.


LucaQuacquarelli:
Holders info: this token is for personal testing purposes. No real circulation is expected.

PR #31249, opened 2025-04-26
Add token 

AdamGabbie:


AdamGabbie:
Fixed :
https://x.com/weloveetits
One Tag : defi  added ,  updated json file .

PR #31313, opened 2025-05-01
Add DSY token info and logo

yamzoni:
This pull request adds DSY Token (0x2d8336471fbae50405ee5888ae927ffc036df84d) info.json and logo.png for TrustWallet integration.


yamzoni (commented):
Updated info.json and added social links as requested.

PR #31609, opened 2025-05-18
Add info.json for Persian Gulf - Shakhab Pars token

Halireza4074:
Add PGSP token info


Halireza4074:
> Add PGSP token info

PR #31942, opened 2025-06-05
Add SkillX token logo and info.json

0RavenPride:


0RavenPride:
the currect is Icone SKX.png

PR #32091, opened 2025-06-13
Add Mgc31 BEP20 token with logo and info.json

itsmgc31:
Add Mgc31 BEP20 token with logo and info.json


itsmgc31:
Why should I pay to add my token in trust wallet? It should be free.

PR #32110, opened 2025-06-15
Updated json file for Polygon active token - Yield90

Yield90:


Yield90:
You are asking me to use https://etherscan.io/token/0xB6F73DDaFa36c3921f3883C1F1c50f69BD74Be05, not https://polygonscan.com/token/0xB6F73DDaFa36c3921f3883C1F1c50f69BD74Be05. My token is deployed on Polygon so that does not work. Also I thought the service was free

PR #32204, opened 2025-06-21
Add FROGEX token info and logo

Heeero3k:
Add FROGEX token info and logo


Heeero3k:
> Add FROGEX token info and logo

PR #32208, opened 2025-06-21
Add FROGEX token info and logo

Heeero3k:
Adding token info.json and logo.png for FROGEX token on Binance Smart Chain.


Heeero3k (commented):
revert accidental deletions in info folder

PR #33024, opened 2025-08-06
Add jbc token logo and metadata

cryptexperiments:


cryptexperiments:
Note: I don’t have permission to add labels, but this is a BEP20 token on BSC with 69,619 holders. Please advise if anything else is required before payment.

PR #33064, opened 2025-08-07
Corwtoken patch 1

corwtoken:
Add CORW Token logo and info.json


corwtoken:
Dear TrustWallet team.
Thanks for your reply. But as I checked, merge for the projects are a good
status are free and our project is a standard with all required and basic
requirements.  Whay should I pay?

PR #33410, opened 2025-09-03
Add OilXCoin (ERC20)

OilXCoin:
Adding info for ERC20 token 'OilXCoin'.


OilXCoin:
Hi Team!

The reported error about invalid links.name does not seem correct. Could it be related to the recent renaming of Twitter -> X?

Regarding the circulation requirement:
We are a Security Token Issuer. This means that only a few licensed exchanges can trade our token. Unlike utility tokens that can be listed on many decentralized exchanges, reaching 10,000 holders is much harder in our regulated environment.

Could you clarify if Security Tokens are treated differently in this case? In particular, we would like to know if paying the Pull Request Fee makes sense - or if the request will be rejected anyway because the holders threshold is not met.

Thank you!
OilXCoin Team

PR #35154, opened 2026-01-16
Add JPYR logo

adachicrypto-debug:


adachicrypto-debug:
JPYR is already listed on BTSE exchange. We are providing official information and logo for Trust Wallet integration. Please check the updated info.json which is now correctly formatted.

PR #35343, opened 2026-02-04
add fomolt to base

fomolt-app:


fomolt-app:
We have 900 holders currently, can wait until we have more.

PR #36900, opened 2026-05-27
Add PAXI token logo

wunmiade69-sudo:
Token name: PAXI
Network: BNB Smart Chain
Contract: 0xeda04581461bc308e60052fe489e20c3f78d6e4e
Logo added in correct folder


wunmiade69-sudo:
It is too cost , listing on exchanges has not begun , because you want to
add picture you want to collect 500twt.
 It is better to wait until I list in major exchanges

PR #37080, opened 2026-06-15
Add Claude Coin (CC) logo - Polygon

escaladadavid10-ctrl:


escaladadavid10-ctrl:
Added info.json file. Both logo.png and info.json are now correctly placed.

PR #37344, opened 2026-07-10
Add Layali Global Token (TEK)

Layaliglobal:
Please add the TEK token to the assets list.


Layaliglobal:
I have updated the info.json file with the required links and tags. Please re-check.

PR #37481, opened 2026-07-28
Add token info for XAUs

k0turbo:


Aranmobinarann (commented):
Checksum Verification
The contract address has been thoroughly verified for EIP-55 checksum compliance. This ensures that the address is case-sensitive correct, preventing any potential typing errors or mismatches when users interact with the token. All characters have been cross-checked against the official source to guarantee 100% accuracy

